PR #4, opened 2017-10-10
 Rely on new-style permission name to guard loan-policy settings

MikeTaylor:
Should fix UICIRC-21.


MikeTaylor:
No-one needs to review this; but policy requires it, so I'd be grateful if one of you guys would click the big green button for me.

MikeTaylor:
I agree with @nielserik's comment.

PR #17, opened 2017-11-01
add validation. Refs UICIRC-20

zburke:
There's a function goin' on right here
Some validation to check the due date years
So bring your from-dates and your to-dates too
We gonna validate all those dates for you 

VALLLLLLLLLLLL-I-DA-TION TIME COME ON!


MikeTaylor:
This is the greatest PR description of all time.

PR #22, opened 2017-11-08
Include fixed due date schedules in loan-policy form. Fixes UICIRC-27

zburke:


zburke:
@mkuklis please take a look at the changes here. The stated goal of the ticket is to call `stripes.connect()` on the form in order to retrieve the fixed-due-date schedules, but I also pulled out the extra "delete" buttons and call to `stripesForm()` that are handled by `EntryManager`. Even so, there is still something funky going on with validation and the "Loanable" checkbox, and I get errors when trying to create new policies. 

I don't want to step on your toes if this is a work-in-progress. LMK if I should go ahead and merge this now, or fix these issues on this branch, or wait until the issues are fixed in master.

PR #90, opened 2018-07-23
FOLIO-1219 feedback on save editing loan rules

tdolega:


MikeTaylor:
(Don't forget to fix the ESLint errors in this app. Not in the present PR, though, best do that separately.)

PR #99, opened 2018-08-22
disable toolbar in view mode for staffslips

aditya-matukumalli:
- pass readOnly prop to disable editing
- Use StaffSlipEditor component to render the editor with tokens and without the html
- Display preview modal on clicking the preview button

### DIFF
--- a/settings/StaffSlips/PreviewModal.js
+++ b/settings/StaffSlips/PreviewModal.js
@@ -26,21 +26,20 @@ class PreviewModal extends React.Component {
     const processNodeDefinitions = new HtmlToReact.ProcessNodeDefinitions(React);
 
     this.demoData = {
-      'Item title': 'Hello',
+      'Item title': '',
       'Item barcode': '<Barcode/>',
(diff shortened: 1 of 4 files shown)


mkuklis:
@aditya-matukumalli this looks good but is there a reason you are using the full editor for the read only view here? Is there anything else you need from the editor besides just displaying the html?

If not could we do something similar to https://github.com/folio-org/ui-circulation/blob/master/settings/StaffSlips/PreviewModal.js and use `HtmlToReact` instead?

aditya-matukumalli:
@mkuklis Thats a good point. I used the StaffSlipEditor component initially just so to make use of the read-only prop on react-quill, which serves the purpose here. But, HtmlToReact definitely looks like a better option as it is a light weight module which serves similar purpose. I made the necessary changes and some slight tweaks to the PreviewModal as well.

mkuklis:
@aditya-matukumalli thank you for refactoring it. It looks great! Also thank you for cleaning up the `PreviewModal`

PR #100, opened 2018-08-23
View and edit mode display for transit slip, refs UICIRC-76

aditya-matukumalli:


settings/StaffSlips/StaffSlipDetail.js
```diff
@@ -32,15 +32,28 @@ class StaffSlipDetail extends React.Component {
...
-   this.detailFormat = {
-      'Item title': '',
-      'Item barcode': '',
-      'Item call number': ' Call number:',
-      'Requester last name': ' Requester Last name:',
-      'Requester first name': ' Requester First name:',
-      'Transaction Id': ' Transaction Id:',
-      'Hold expiration': ' Hold expiration:',
-    };
+   this.previewFormat = props.initialValues.name === 'Hold' ?
```
mkuklis: @aditya-matukumalli it looks like the same structure is repeated in `PreviewModal` and `StaffSlipDetail`. Do you think we could perhaps extract it into a separate file (perhaps `formats.js`).
It could be: 
```js
const formats = {
  Hold: {
  ...
  },
  Transit: {
  ...
  }
};
```

Then we could just do:

```js
this.previewFormat = formats[props.initialValues.name];
```

would that work?

aditya-matukumalli: @mkuklis  Good point. Will do

mkuklis: @aditya-matukumalli thank you!

PR #103, opened 2018-08-28
Fix mustache format plus cleanup. Part of UICIRC-83

mkuklis:
This PR fixes the mustache formatting from `{`, `}`  to `{{`, `}}` plus cleans up the editor code a bit.


mkuklis:
@aditya-matukumalli  I think I broke https://issues.folio.org/browse/UICIRC-79 with this PR. Could you tell me what should be displayed on the details screen? It's not clear to me what is expected when I read UICIRC-79. Should the formatting be preserved but instead of the preview values we should show tokens?

PR #115, opened 2018-10-03
Use stripes 1.0

cherewaty:
Uses new `@folio/stripes` package.


cherewaty:
Outstanding issue: imports form css from `stripes-components`.

PR #120, opened 2018-10-30
[UICIRC-91] Use documented react-intl patterns instead of stripes.intl

anastasiaHahanova:
## Purpose
Taking into account that retrieving the intl object from this.context.intl or the stripes god object (stripes.intl) is a deprecated pattern.

The purpose of this PR is to update this repo to only use react-intl constructs outlined at https://github.com/folio-org/stripes-core/blob/master/doc/i18n.md

## Learning
React Intl docs: https://github.com/yahoo/react-intl/wiki

### DIFF
--- a/settings/FixedDueDateScheduleDetail.js
+++ b/settings/FixedDueDateScheduleDetail.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, intlShape, injectIntl } from 'react-intl';
 import { Accordion, Col, ExpandAllButton, KeyValue, Row } from '@folio/stripes/components';
 import { stripesShape } from '@folio/stripes/core';
 import { ViewMetaData } from '@folio/stripes/smart-components';
@@ -10,6 +10,7 @@ import css from './FixedDueDateSchedule.css';
 class FixedDueDateScheduleDetail extends React.Component {
   static propTypes = {
     initialValues: PropTypes.object,
+    intl: intlShape.isRequired,
     stripes: stripesShape.isRequired,
   }
 
@@ -17,7 +18,6 @@ class FixedDueDateScheduleDetail extends React.Component {
     super(props);
     this.handleSectionToggle = this.handleSectionToggle.bind(this);
     this.handleExpandAll = this.handleExpandAll.bind(this);
-    this.formatDate = props.stripes.formatDate;
     this.cViewMetaData = props.stripes.connect(ViewMetaData);
     this.state = {
       sections: {
@@ -44,8 +44,11 @@ class FixedDueDateScheduleDetail extends React.Component {
   }
 
   render() {
-    const fixedDueDateSchedule = this.props.initialValues;
-    const formatMsg = this.props.stripes.intl.formatMessage;
+    const {
+      intl: { formatDate, formatMessage },
+      initialValues: fixedDueDateSchedule,
+    } = this.props;
+
     const { sections } = this.state;
     const renderSchedules = fixedDueDateSchedule.schedules.map((schedule, index) => (
       <div key={index} className={css.scheduleItem}>
@@ -69,9 +72,9 @@ class FixedDueDateScheduleDetail extends React.Component {
             </Col>
           </Row>
           <Row key={index + 2}>
-            <Col xs={4}>{this.formatDate(schedule.from)}</Col>
-            <Col xs={4}>{this.formatDate(schedule.to)}</Col>
-            <Col xs={4}>{this.formatDate(schedule.due)}</Col>
+            <Col xs={4}>{formatDate(schedule.from)}</Col>
+            <Col xs={4}>{formatDate(schedule.to)}</Col>
+            <Col xs={4}>{formatDate(schedule.due)}</Col>
           </Row>
         </div>
       </div>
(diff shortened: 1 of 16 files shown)


settings/FixedDueDateScheduleDetail.js
```diff
@@ -69,9 +72,9 @@ class FixedDueDateScheduleDetail extends React.Component {
             </Col>
           </Row>
           <Row key={index + 2}>
-            <Col xs={4}>{this.formatDate(schedule.from)}</Col>
-            <Col xs={4}>{this.formatDate(schedule.to)}</Col>
-            <Col xs={4}>{this.formatDate(schedule.due)}</Col>
+            <Col xs={4}>{formatDate(schedule.from)}</Col>
```
Godlevskyi: Please use \<FormattedDate\> component here

settings/FixedDueDateScheduleDetail.js
```diff
@@ -44,8 +44,11 @@ class FixedDueDateScheduleDetail extends React.Component {
   }
 
   render() {
-    const fixedDueDateSchedule = this.props.initialValues;
-    const formatMsg = this.props.stripes.intl.formatMessage;
+    const {
+      intl: { formatDate, formatMessage },
```
Godlevskyi: No need to use formatDate, please import <formatedDate> component

settings/FixedDueDateScheduleForm.js
```diff
@@ -143,28 +154,34 @@ Delete
   }
 
   renderPaneTitle() {
-    const { initialValues, stripes: { intl } } = this.props;
+    const {
+      initialValues,
```
Godlevskyi: InitialValues direcly reassigned after destructuring. So, it make sence to reassign in the destructuring and  also set default value

settings/FixedDueDateScheduleForm.js
```diff
@@ -143,28 +154,34 @@ Delete
...
+      intl: { formatMessage },
+    } = this.props;
+
     const selectedSet = initialValues || {};
```
Godlevskyi: Should be removed. Se comment above

settings/LoanPolicyForm.js
```diff
@@ -50,7 +51,7 @@ class LoanPolicyForm extends React.Component {
   }
 
   getCurrentValues() {
-    const { stripes: { store } } = this.props;
+    const { store } = this.props.stripes;
```
Godlevskyi: What is the reason for that change?

settings/LoanRules.js
```diff
@@ -21,6 +22,7 @@ const editorDefaultProps = {
     m: 'Material Type',
     t: 'Loan Type',
   },
+  intl: intlShape.isRequired,
```
Godlevskyi: ?? what for it is nat a propTypes

settings/LoanRules.js
```diff
@@ -162,7 +180,12 @@ class LoanRules extends React.Component {
   }
 
   render() {
-    if (!this.props.resources.loanTypes) {
+    const {
+      resources,
+      intl: { formatMessage },
+    } = this.props;
+
+    if (!resources.loanTypes) {
```
Godlevskyi: Only loanTypes used from resources, so it could be movewd to destructuring above, like:
resources: { loanTypes }

settings/StaffSlips/PreviewModal.js
```diff
@@ -39,9 +39,15 @@ class PreviewModal extends React.Component {
...
-    const closeLabel = intl.formatMessage({ id: 'ui-circulation.settings.staffSlips.close' });
-    const heading = intl.formatMessage({ id: 'ui-circulation.settings.staffSlips.previewLabel' });
+    const {
+      open,
+      onClose,
+      previewTemplate,
+      intl: { formatMessage },
+    } = this.props;
+
+    const closeLabel = formatMessage({ id: 'ui-circulation.settings.staffSlips.close' });
```
Godlevskyi: Please use \<FormatedMessage\> here as string is not required

settings/LoanPolicyForm.js
anastasiaHahanova: I see no need for nested destructuring in this case. Unlikely some other properties of props will be used in this method, therefore, now it looks more readable.

settings/FixedDueDateScheduleForm.js
```diff
@@ -143,28 +154,32 @@ Delete
   }
 
   renderPaneTitle() {
-    const { initialValues, stripes: { intl } } = this.props;
-    const selectedSet = initialValues || {};
+    const {
+      initialValues: selectedSet = {},
```
mkuklis: I think this will override the `initialValues` which changes the previous implementation. Could you please change it back to what it was? or just do:

`const selectedSet = this.props.initialValues || {}`

We have to be careful not to mutate `initialValues`. redux-form is sensitive about that object.

settings/StaffSlips/StaffSlipDetail.js
```diff
@@ -52,7 +53,7 @@ class StaffSlipDetail extends React.Component {
 
   render() {
     const { openDialog } = this.state;
-    const staffSlip = this.props.initialValues;
+    const { initialValues: staffSlip } = this.props;
```
mkuklis: This also can't be right. Could you please revert this change?

settings/FixedDueDateScheduleForm.js
```diff
@@ -109,24 +114,30 @@ class FixedDueDateScheduleForm extends React.Component {
...
             <Button
               id="clickable-delete-set"
-              title={intl.formatMessage({ id: 'ui-circulation.settings.fDDSform.delete' })}
+              title={formatMessage({ id: 'ui-circulation.settings.fDDSform.delete' })}
               buttonStyle="danger"
               onClick={this.beginDelete}
               disabled={confirmDelete}
             >
-Delete
+              Delete
```
mkuklis: @anastasiaHahanova would you mind also translating this one? Thank you!

settings/FixedDueDateScheduleForm.js
anastasiaHahanova: Hi Michał, you probably haven't mentioned that in this particular case I have applied ES6 object destructuring pattern. You can have a look of the docs [here](https://javascript.info/destructuring-assignment). In this case I believe we don't need to create a new variable as we can distructure it directly from this.props object with the default value.

mkuklis: Oh you are right @anastasiaHahanova! Sorry about that. Somehow it was hard spot that for me.

settings/StaffSlips/StaffSlipDetail.js
mkuklis: This of course is also fine.

PR #125, opened 2018-11-07
UICIRC-91 Replace formatMessage() with FormattedMessage

cherewaty:
## Purpose
- Same pattern as https://github.com/folio-org/ui-checkout/pull/201
- Related to https://github.com/folio-org/stripes-components/pull/696 and https://github.com/folio-org/stripes-smart-components/pull/331

## Approach
- Exclusively use `<FormattedMessage>`, `<FormattedDate>`, and `<FormattedTime>` from `react-intl`. 
- Discontinue using `intl.formatMessage()`, `intl.formatDate()`, and `intl.formatTime()`.

### DIFF
--- a/settings/FixedDueDateScheduleForm.js
+++ b/settings/FixedDueDateScheduleForm.js
@@ -103,14 +103,17 @@ class FixedDueDateScheduleForm extends React.Component {
 
     return (
       <PaneMenu>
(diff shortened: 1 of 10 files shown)


mkuklis:
Hey @cherewaty would it be possible to merge this one in?

PR #131, opened 2018-11-20
UICIRC-64. Opening time offset field added. Components refactored.

anton-plotnikov:
Opening Time offset fields added to Loan policy in scope of UICIRC-64

### DIFF
--- a/settings/LoanPolicyDetail.js
+++ b/settings/LoanPolicyDetail.js
@@ -1,9 +1,20 @@
-import _ from 'lodash';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+import {
+  get,
+  find,
+  cloneDeep,
+} from 'lodash';
+
 import { stripesShape } from '@folio/stripes/core';
-import { Accordion, Col, ExpandAllButton, KeyValue, Row } from '@folio/stripes/components';
+import {
+  Accordion,
+  ExpandAllButton,
+  KeyValue,
+  Col,
+  Row,
+} from '@folio/stripes/components';
 import { ViewMetaData } from '@folio/stripes/smart-components';
 
 import {
@@ -40,7 +51,7 @@ class LoanPolicyDetail extends React.Component {
 
   handleExpandAll(sections) {
     this.setState((curState) => {
-      const newState = _.cloneDeep(curState);
+      const newState = cloneDeep(curState);
       newState.sections = sections;
       return newState;
     });
@@ -48,7 +59,7 @@ class LoanPolicyDetail extends React.Component {
 
   handleSectionToggle({ id }) {
     this.setState((curState) => {
-      const newState = _.cloneDeep(curState);
+      const newState = cloneDeep(curState);
       newState.sections[id] = !newState.sections[id];
       return newState;
     });
@@ -60,7 +71,7 @@ class LoanPolicyDetail extends React.Component {
       ...longTermLoansOptions,
     ];
 
-    return _.find(availableLoansOptions, loanOption => loanOption.id === selectedItemId);
+    return find(availableLoansOptions, loanOption => loanOption.id === selectedItemId);
   };
 
   renderLoans() {
@@ -82,13 +93,14 @@ class LoanPolicyDetail extends React.Component {
       schedule = fixedDueDateSchedules.find(s => s.id === policy.loansPolicy.fixedDueDateScheduleId);
     }
 
-    const profileId = _.get(policy, ['loansPolicy', 'profileId']);
-    const profile = _.find(loanProfileTypes, t => t.value === profileId);
-    const ddId = _.get(policy, ['loansPolicy', 'closedLibraryDueDateManagementId']);
+    const profileId = get(policy, ['loansPolicy', 'profileId']);
+    const profile = find(loanProfileTypes, t => t.value === profileId);
+    const ddId = get(policy, ['loansPolicy', 'closedLibraryDueDateManagementId']);
     const closedLibraryDueDateManagementItem = this.getClosedLibraryDueDateManagementItem(ddId);
-    const periodInterval = _.get(policy, ['loansPolicy', 'period', 'intervalId']);
-    const exReqPerInterval = _.get(policy, ['loansPolicy', 'existingRequestsPeriod', 'intervalId']);
-    const gracePeriodInterval = _.get(policy, ['loansPolicy', 'gracePeriod', 'intervalId']);
+    const periodInterval = get(policy, ['loansPolicy', 'period', 'intervalId']);
+    const exReqPerInterval = get(policy, ['loansPolicy', 'existingRequestsPeriod', 'intervalId']);
+    const gracePeriodInterval = get(policy, ['loansPolicy', 'gracePeriod', 'intervalId']);
+    const timeOffsetInterval = get(policy, ['loansPolicy', 'openingTimeOffset', 'intervalId']);
 
     return (
       <div>
@@ -104,7 +116,7 @@ class LoanPolicyDetail extends React.Component {
           <Col xs={12}>
             <KeyValue
               label={<FormattedMessage id="ui-circulation.settings.loanPolicy.loanProfile" />}
-              value={_.get(profile, ['label'], '-')}
+              value={get(profile, ['label'], '-')}
             />
           </Col>
         </Row>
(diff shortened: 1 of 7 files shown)


settings/components/PolicyPropertyDescription/PolicyPropertyDescription.js
```diff
@@ -0,0 +1,46 @@
+import React, { Fragment } from 'react';
+import PropTypes from 'prop-types';
+import _ from 'lodash';
```
OleksandrYehorovEPAM: Please consider importing only required lodash methods instead of importing the whole library.
```suggestion
import { get } from 'lodash';
```

settings/LoanPolicyForm.js
```diff
@@ -114,6 +115,10 @@ class LoanPolicyForm extends React.Component {
     if (altRenewPeriod && altRenewPeriod.intervalId && !altRenewPeriod.duration) {
       this.props.change('renewalsPolicy.period.duration', 1);
     }
+    const openingTimeOffset = (allValues.loansPolicy && allValues.loansPolicy.openingTimeOffset);
+    if (openingTimeOffset && openingTimeOffset.intervalId && !openingTimeOffset.duration) {
+      this.props.change('loansPolicy.openingTimeOffset.duration', 0);
```
OleksandrYehorovEPAM: Seems reasonable for me to use destructured "change" function here and above.  
```suggestion
     const { change } = this.props;
     
    ...
     
     change('loansPolicy.openingTimeOffset.duration', 0);
```

settings/LoanPolicyDetail.js
```diff
@@ -2,9 +2,17 @@ import _ from 'lodash';
...
 import { stripesShape } from '@folio/stripes/core';
-import { Accordion, Col, ExpandAllButton, KeyValue, Row } from '@folio/stripes/components';
+import {
+  Accordion,
+  ExpandAllButton,
+  KeyValue,
+  Col,
+  Row,
+} from '@folio/stripes/components';
 import { ViewMetaData } from '@folio/stripes/smart-components';
```
OleksandrYehorovEPAM: There should be an empty line after folio imports in order to separate them from own imports
```suggestion
import { ViewMetaData } from '@folio/stripes/smart-components';

```

anton-plotnikov: Both the imports are from the @folio

OleksandrYehorovEPAM: Isn't PolicyPropertyDescription on line 15 is your own import?

settings/components/PolicyPropertyDescription/PolicyPropertyDescription.js
anton-plotnikov: Tree shaking in Webpack takes place independently of the way we use our imports: named or not. Still, it is a question, is the tree-shacking going to take place or not. The thing is we are using lodash library, not lodash-es and the former does not decalre any named export, so I assume that using named exports will not give us any performance boost.
Correct me if I'm wrong, please

settings/LoanPolicyDetail.js
anton-plotnikov: Apologies, missed that

settings/components/PolicyPropertyDescription/PolicyPropertyDescription.js
OleksandrYehorovEPAM: Sounds reasonable about tree shaking so it is just about code style preferences. I see in this repository lodash imports like yours are being used so I assume it would be fine to leave it as is.

settings/components/PolicyPropertyDescription/PolicyPropertyDescription.js
```diff
@@ -0,0 +1,46 @@
...
+
+import {
+  Col,
+  Row,
+  KeyValue,
+} from '@folio/stripes/components';
+
+const PolicyPropertyDescription = ({
+  policy,
```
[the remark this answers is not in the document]
anton-plotnikov: Seems that the logic becomes more complicated, really. Totally agree

settings/components/PolicyPropertySetter/PolicyPropertySetter.js
```diff
@@ -0,0 +1,80 @@
...
+  textFieldName = loanHeader,
+  selectFieldName = loanHeader,
+  intervalPeriods,
+  validator,
+  loansPolicyNamespace,
+  loanHeaderNamespace,
+  durationDescriptor,
+  timeIntervalsUnitsDescriptor,
+}) => (
+  <Fragment>
```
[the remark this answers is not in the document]
anton-plotnikov: The div wrapper is redundant, sure. Going to leave only the <Fragment>

settings/components/PolicyPropertySetter/PolicyPropertySetter.js
```diff
@@ -0,0 +1,80 @@
...
+    <div>
+      <p>
```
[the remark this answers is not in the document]
anton-plotnikov: Using `<Row>` element will break the layout and in this case we need to add a `className` property and additional stylesheet for this case. Wouldn't it be better to leave the `<p>` tag?

settings/components/PolicyPropertySetter/PolicyPropertySetter.js
```diff
@@ -0,0 +1,80 @@
...
+  Row,
+  Select,
+  TextField,
+} from '@folio/stripes/components';
+
+// If text and select fields names' are not specified,
+// the header is used to name these fields
+const PolicyPropertySetter = ({
+  loanHeader,
+  textFieldName = loanHeader,
```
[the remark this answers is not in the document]
anton-plotnikov: Completely agree, cause they may change independently

settings/LoanPolicyForm.js
```diff
@@ -26,6 +27,7 @@ import {
   loanProfileMap,
   shortTermLoansOptions,
   longTermLoansOptions,
+  BEGINNING_OF_THE_NEXT_OPEN_SERVICE_POINT_HOURS as timeOffsetLabel,
```
maximdidenkoepam: I told you not to use variable renaming. In the method below you compare id with label

settings/LoanPolicyDetail.js
```diff
@@ -109,17 +117,17 @@ class LoanPolicyDetail extends React.Component {
           </Col>
         </Row>
         {policy.loansPolicy.profileId === loanProfileMap.ROLLING &&
-          <div>
```
maximdidenkoepam: if you just reverted changes why are they present in this file as new code?

anton-plotnikov: Missed some spaces to set

settings/LoanPolicyForm.js
```diff
@@ -114,6 +116,10 @@ class LoanPolicyForm extends React.Component {
     if (altRenewPeriod && altRenewPeriod.intervalId && !altRenewPeriod.duration) {
       this.props.change('renewalsPolicy.period.duration', 1);
     }
+    const openingTimeOffset = (allValues.loansPolicy && allValues.loansPolicy.openingTimeOffset);
+    if (openingTimeOffset && openingTimeOffset.intervalId && !openingTimeOffset.duration) {
```
VictorSoroka1: Move condition inside if to the descriptive variable and add an empty line before and after `const`.

settings/LoanPolicyForm.js
```diff
@@ -164,6 +170,17 @@ class LoanPolicyForm extends React.Component {
         </FormattedMessage>
       ));
 
+    const setTimeOffsetVisibility = () => {
```
VictorSoroka1: Seems like the name of the function does not fully reflect the meaning of the function. The function itself does not `set` anything. Rather it returns something. Should it start with `get`?

settings/LoanPolicyForm.js
```diff
@@ -1,23 +1,24 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Field, getFormValues } from 'redux-form';
-import _ from 'lodash';
-
 import { FormattedMessage } from 'react-intl';
+import _ from 'lodash';
```
mkuklis: @anton-plotnikov would it be ok if you just import what you need from lodash instead of the full `_` ( I think we can do the same in `LoanPolicyDetail.js`)? This should help with tree shaking in the future.

PR #176, opened 2019-01-14
UICIRC-151: Add validation rules.

OleksandrAntonenko1:
# Description
Purpose: To require staff users to select values for certain fields under various conditions.

Story: As a staff member, I want to be reminded that certain fields are conditionally required before saving and closing the Loan Policy Editor when adding or editing loan policies.

Scenario 1:

Issue 1: Add alert for required field - Fixed due date schedule
For Loan Policy where:

Loanable = Y
Loan type = Fixed
'Fixed due date schedule' should be a required field
Else: Flag field as required when user clicks 'Save and close' button. Alert should read "! Please make a selection" (note: an existing convention exists for non-select fields - "! Please fill out this field", which could be used if necessary)
Issue 2: Add alert for required field - Alternate fixed due date schedule for renewals
For Loan Policy where:

Loanable = Y
Loan type = Fixed
Renewable = Y
Renewal period different from original loan = Y
'Alternate fixed due date schedule for renewals' should be a required field
Else: Flag field as required when user clicks 'Save and close' button. Alert should read "! Please make a selection" (note: an existing convention exists for non-select fields - "! Please fill out this field", which could be used if necessary)
Issue 3: Add alert for required field - Alternate schedule for renewals
For Loan Policy where:

Loanable = Y
Loan type = Rolling
Renewable = Y
Renewal period different from original loan = Y
'Alternate loan period for renewals' should be a required field
Else: Flag field as required when user clicks 'Save and close' button. Alert should read "! Please fill out this field"
# Screenshots
## Loan type = Fixed
screen shot 2019-01-14 at 4 13 39 pm
## Loan type = Rolling
screen shot 2019-01-14 at 4 13 56 pm


settings/Validation/FormValidator.js
```diff
@@ -15,6 +15,10 @@ const defaultValidators = {
     validate: (value) => isNumber(value) || !isEmpty(value),
     message: <FormattedMessage id="ui-circulation.settings.validate.fillIn" />,
   },
+  isNotEmptySelect: {
+    validate: (value) => isNumber(value) || !isEmpty(value),
```
maximdidenkoepam: Please use validators composition for this case

OleksandrAntonenko1: Done.

mkuklis:
Looks good @OleksandrAntonenko1!

PR #183, opened 2019-01-18
UICIRC-98 - Add new/edit notice policy form

maximdidenkoepam:
# Purpose
Library staff (administrators) need to be able to setup and configure notice policies to define triggering events and conditions and utilize notice templates

# Link
https://issues.folio.org/browse/UICIRC-98

# Screenshots
screen shot 2019-01-18 at 11 47 01
screen shot 2019-01-18 at 11 49 07


settings/LoanPolicy/components/DeleteEntry/DeleteEntry.js
```diff
@@ -0,0 +1,60 @@
...
+
+import { DeleteConfirmationModal } from '../../../components';
+
+class DeleteEntity extends React.Component {
+  static propTypes = {
+    isOpen: PropTypes.bool.isRequired,
+    perm: PropTypes.string.isRequired,
+    policyName: PropTypes.string.isRequired,
+    deleteEntityKey: PropTypes.string.isRequired,
+    initialValues: PropTypes.object,
```
OleksandrAntonenko1: Please add isRequired to errors property or defaultProps if this property is not required

settings/NoticePolicy/NoticePolicyForm.js
```diff
@@ -1,9 +1,161 @@
...
+  GeneralSection,
+  LoanNoticesSection,
+  FeeFineNoticesSection,
+  RequestNoticesSection,
+} from './components';
 
 class NoticePolicyForm extends React.Component {
+  static propTypes = {
+    stripes: stripesShape.isRequired,
+    pristine: PropTypes.bool,
```
OleksandrAntonenko1: Please add isRequired to errors property or defaultProps if this property is not required

settings/NoticePolicy/components/EditSections/GeneralSection/GeneralSection.js
```diff
@@ -0,0 +1,70 @@
...
+import { FormattedMessage } from 'react-intl';
+
+import {
+  TextArea,
+  TextField,
+  Checkbox,
+  Accordion,
+} from '@folio/stripes/components';
+
+import { Metadata } from '../../../../components';
```
OleksandrAntonenko1: Can we use import with the alias to make the path shorter and more clear

settings/Validation/NoticePolicy.js
```diff
@@ -1,3 +1,12 @@
-export default function () {
-  return true;
+import FormValidator from './FormValidator';
+
+export default function (policy) {
+  const config = {
+    'name': {
```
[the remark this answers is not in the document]
maximdidenkoepam: In this case it is not obvious, but key should be a string because it can be something like `'noticePolicies.period.name'` and for consistency in this file I suggest to use strings as object keys.

PR #192, opened 2019-01-24
be lenient with react deps

zburke:
Workspace builds are all kinds of whacked. It appears that, maybe, we
are getting conflicting versions of react in the bundle because of ~
deps on a version that is older than the one provided by ^ deps from
other modules, maybe? It's not totally clear. But using a ^ dep here
will at least be consistent with other apps. Consistency is good.

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,8 +150,8 @@
     "babel-polyfill": "^6.26.0",
     "chai": "^4.2.0",
     "eslint": "^5.5.0",
-    "react": "~16.7.0",
-    "react-dom": "~16.6.0",
+    "react": "^16.7.0",
+    "react-dom": "^16.6.0",
     "react-redux": "~5.1.1",
     "redux": "~4.0.0",
     "sinon": "^7.2.2"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -150,8 +150,8 @@
     "babel-polyfill": "^6.26.0",
     "chai": "^4.2.0",
     "eslint": "^5.5.0",
-    "react": "~16.7.0",
-    "react-dom": "~16.6.0",
+    "react": "^16.7.0",
```
ttamj: Is there anything from `16.7` that ui-circulation depends on?  Looks like `stripes` and `platform-core` are requesting `~16.6` which ui-circulation will complain about during a platform install since its less than the minor version requested here.

mkuklis: @ttamj that's what started causing issues for me. I'm not sure if `16.7` is needed here.

OleksandrAntonenko1: @ttamj It is not necessary. I just got a problem with duplication of React and I have fixed it in such a way and probably forgot to delete this dependency. So just feel free to revert it to `~16.6`

PR #211, opened 2019-01-31
use standard repository structure

zburke:
Fixes [UICIRC-146](https://issues.folio.org/browse/UICIRC-146)


zburke:
Replaced by #251

PR #223, opened 2019-02-07
UICIRC-106 - Single patron notice within a policy

maximdidenkoepam:
# Purpose
To add ability to create loan notices for notice policy.

# Link
https://issues.folio.org/browse/UICIRC-106

# Screenshots
screen shot 2019-02-07 at 16 20 55
screen shot 2019-02-07 at 16 21 12

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -102,4 +102,36 @@ export const requestPolicyTypes = [
   'Recall',
 ];
 
+export const loanNoticesFormats = [
+  { value: 'Email', label: 'Email' },
+];
+
+export const loanNoticesFrequencyMap = {
+  ONE_TIME: 'One time',
+  RECURRING: 'Recurring',
+};
+
+export const loanNoticesSendEventMap = {
+  AFTER: 'After',
+  BEFORE: 'Before',
+  UPON: 'Upon At'
+};
+
+export const loanNoticesFrequency = [
+  { value: loanNoticesFrequencyMap.ONE_TIME, label: loanNoticesFrequencyMap.ONE_TIME },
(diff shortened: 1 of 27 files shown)


constants.js
```diff
@@ -102,4 +102,36 @@ export const requestPolicyTypes = [
...
+];
+
+export const loanNoticesSendEvent = [
+  { value: loanNoticesSendEventMap.UPON, label: 'Upon/At' },
+  { value: loanNoticesSendEventMap.BEFORE, label: loanNoticesSendEventMap.BEFORE },
+  { value: loanNoticesSendEventMap.AFTER, label: loanNoticesSendEventMap.AFTER },
+];
+
+export const loanNoticesSendWhen = [
+  { value: 'Due date', label: 'Due date' },
```
mkuklis: @maximdidenko I think we will need to translate the labels here right?

constants.js
```diff
@@ -102,4 +102,36 @@ export const requestPolicyTypes = [
...
+  { value: 'Overdue', label: 'Overdue' },
```
mkuklis: @maximdidenkoepam same here for label.

settings/NoticePolicy/NoticePolicyForm.js
```diff
@@ -45,12 +50,28 @@ class NoticePolicyForm extends React.Component {
...
     sections: {
       general: true,
-      loanNotices: false,
+      loanNotices: true,
       feeFineNotices: false,
       requestNotices: false,
     },
   };
 
+  getTemplates = (noticeCategory) => {
```
mkuklis: It looks like `getTemplates` exists in both `NoticePolicyForm` and `NoticePolicyDetail`. Perhaps it would be valuable to extract it from the components into a `util` file so they can be reused.

PR #256, opened 2019-02-20
Do not show unsaved changes popup when policy form is not dirty. Fixes UICIRC-104

mkuklis:
It looks like the `change` would execute in cases when the default values were missing or in some cases wrong which could cause the form to get "dirty". This PR fixes it.

More info:
https://issues.folio.org/browse/UICIRC-104


mkuklis:
Thank you for catching this @maximdidenko. 

I moved this to the LoanPolicy model because in the previous implementation if the `setDueDateManagementSelectedId` was set to incorrect value (for example on the server) or if it was empty the `change` would propagate and make the form `dirty` which is not what we want. The form needs to stay pristine until the user changes something manually. 

I will try to fix the issue you found. I think we may need to have that check in both places:

1. Set correct value in LoanPolicy model when the form is being initialized
2. Listen for changes in `componentDidUpdate` and update accordingly

PR #257, opened 2019-02-21
UICIRC-190: Add tests for Loan policy

OleksandrAntonenko1:


test/bigtest/network/config.js
```diff
@@ -333,16 +266,66 @@ export default function config() {
     ]
   });
 
+  this.get('/patron-notice-policy-storage/patron-notice-policies', {
```
mkuklis: @OleksandrAntonenko1 could you please create a factory for this. The default scenario could be then used to populate couple of the notice policies. My understanding is that it's better to rely on actual factories and not hard coded values.  I think we have other hard coded responses here. I think we should probably also clean those perhaps in another PR.

test/bigtest/network/scenarios/stabbedLoanPolicies.js
```diff
@@ -0,0 +1,68 @@
+export default (server) => {
+  server.get('/loan-policy-storage/loan-policies', {
```
mkuklis: I don't think this belongs to scenario. Shouldn't this be defined in `config.js`? Also the factory should be used for generating them. Then we can use the scenario to create a couple of these records in a similar fashion to: https://github.com/folio-org/ui-checkin/blob/master/test/bigtest/network/scenarios/default.js#L4-L5

test/bigtest/network/config.js
OleksandrAntonenko1: I think the factory for notice-policy will be added in the scope of another story, which is already in the upcoming sprint.

test/bigtest/network/scenarios/stabbedLoanPolicies.js
OleksandrAntonenko1: Done.

PR #268, opened 2019-02-27
correctly format the dataOptions for a Select

zburke:
`<Select>` accepts a `dataOptions` prop that expects to be an array of
key-value pairs with the props `value` and `label`. It was receiving,
instead, an array of `<option>` elements which caused rendered list to
show up as a series of empty `<option />`s. TBH, I'm not sure how this
ever worked.

Refs [STRIPES-602](https://issues.folio.org/browse/STRIPES-602)


mkuklis:
@zburke nice find!

PR #275, opened 2019-03-11
UICIRC-211 - changed path for patron notice template subject field

maximdidenkoepam:
# Purpose 
Change path for template subject field

# Link
https://issues.folio.org/browse/UICIRC-211

### DIFF
--- a/src/settings/PatronNotices/PatronNoticeDetail.js
+++ b/src/settings/PatronNotices/PatronNoticeDetail.js
@@ -106,7 +106,7 @@ class PatronNoticeDetail extends React.Component {
               <div>
                 <Row>
                   <Col xs={8}>
-                    <KeyValue label={<FormattedMessage id="ui-circulation.settings.patronNotices.subject" />} value={notice.subject} />
+                    <KeyValue label={<FormattedMessage id="ui-circulation.settings.patronNotices.subject" />} value={notice.localizedTemplates.en.header} />
                   </Col>
                   <Col xs={4}>
                     <Button onClick={this.openPreviewDialog}><FormattedMessage id="ui-circulation.settings.patronNotices.preview" /></Button>
--- a/src/settings/PatronNotices/PatronNoticeForm.js
+++ b/src/settings/PatronNotices/PatronNoticeForm.js
@@ -263,7 +263,7 @@ class PatronNoticeForm extends React.Component {
                   <Col xs={8}>
                     <Field
                       label={<FormattedMessage id="ui-circulation.settings.patronNotices.subject" />}
-                      name="subject"
+                      name="localizedTemplates.en.header"
                       id="input-patron-notice-subject"
                       component={TextField}
                     />


src/settings/PatronNotices/PatronNoticeDetail.js
```diff
@@ -106,7 +106,7 @@ class PatronNoticeDetail extends React.Component {
               <div>
                 <Row>
                   <Col xs={8}>
-                    <KeyValue label={<FormattedMessage id="ui-circulation.settings.patronNotices.subject" />} value={notice.subject} />
+                    <KeyValue label={<FormattedMessage id="ui-circulation.settings.patronNotices.subject" />} value={notice.localizedTemplates.en.header} />
```
zburke: Neither the old key, `notice.subject`, nor the new one, `notice.localizedTemplates.en.header` are present in the translations file `en.json`. 

Can you explain the presence of `.en.` in the new key? That looks suspiciously like a locale-related value, which seems incorrect.

maximdidenkoepam: @zburke it's not a key, it's a path to a value.

maximdidenkoepam: At the moment ui does not have an ability to choose language for template. The main reason why it was done is because subject saved to incorrect field and BE guys asked us to set correct path. Body field also saved in `notice.localizedTemplates.en`.

zburke: Oh; my bad. I mis-read it as the `FormattedMessage` key. 

I'll follow up on UICIRC-211 directly. The presence of {{.en.}} in the path still seems odd to me.

zburke: But `en` is not even a legit-locale; we break it down into `en-GB`, `en-SE`, and `en-US`. I am concerned that we're codifying an approach to managing localization that hasn't actually been decided upon and, by coding an invalid locale directly into the source, we are actually making the future work more difficult than it needs to be.

maximdidenkoepam: At the moment `en` is used in password management flow and changing this may take a lot of time, and we don't have it. This approach was chosen before I changed the field path and we need this functionality in order to send loan notices.

dltj: I'm trying to understand why there is a locale label in the key name.  Is there an issue in Jira where the need to support translated templates?  Hard coding the locale label here seems like the wrong approach.

kostyantyn-kh: templates functionality was implemented in scope of this story
https://issues.folio.org/browse/MODLOGIN-56
as far as i understand templates have "language" not "locale" to have an ability to send templates in different languages (not implemented yet)
sorry for misunderstanding

zburke: Looking more closely at the implementation in MODLOGIN-56, I am actually more comfortable with this PR as it shows the key `notice.localizedTemplates.en` actually resolves/translates to a structure like 
```
notice: {
    templates: [
        { lang: "en", template: "some thing" ...}
        { lang: "fr", template: "autre chose" ...}
...
``` 
IOW, this is really "we're harding coding english as the sole member of a field that correctly supports multiple members". 

... But I still think `en` is the wrong key; shouldn't it be `en-US`?

kostyantyn-kh: @zburke need to think about this key in future
currently internal logic relies on key "en", too risky to change it now

maximdidenkoepam: We are going to merge this PR in order to complete sending loan notices till Friday (Discussed with Khalilah Gambrell). Please create separate story with discussed requirements in order to track the issue.

PR #283, opened 2019-03-14
Increase loan policy limit

skomorokh:
While it would be ideal to implement pagination / infinite scroll here, that currently depends on the `EntryManager` component which is due to be refactored already so for now I've increased the maximum number of loan policies / fixed due date schedules returned in the first request to 100 from the default 10.


OleksandrAntonenko1:
I don't think that it is a solid solution. But I agree to do it in such a way until the entry manager would be changed.

PR #286, opened 2019-03-18
UICIRC-160: Change Patron Notice Description field.

OleksandrAntonenko1:
# Description
## Steps to repro:

1. Log into folio_snapshot
2. Go to Settings > Circulation > Patron Notices
3. Click to create a new notice
4. Attempt to save without specifiying a Patron notice description
5. Note Patron notice description is required
**Expected**: Description doesn't seem like it should be a required field. It isn't required in other forms (e.g. loan policies, fixed due date schedules, locations). If it is supposed to be required, it should have an asterisk to indicate it is required (see UICIRC-159 for similar bug)

**Actual**: Description is required


maximdidenkoepam:
@OleksandrAntonenko1 please add to description, that before merging this task BE validation should be also removed.

PR #292, opened 2019-03-21
Eject from settings smart component UICIRC-201

skomorokh:
This doesn't fully implement the new routing architecture but takes a solid step in that direction. Based on Jeffrey's previous work: 

https://github.com/folio-org/ui-circulation/pull/184/


src/settings/index.js
```diff
@@ -1,89 +1,87 @@
...
+            </IfPermission>
+            <IfPermission perm="ui-circulation.settings.request-policies">
+              <NavListItem to={`${path}/request-policies`}>
+                <FormattedMessage id="ui-circulation.settings.index.requestPolicies" />
+              </NavListItem>
+            </IfPermission>
+          </NavList>
+        </Pane>
+        {children}
+      </Fragment>
```
JohnC-80: I feel like we could consolidate some of the repetition here if we stash the routespaths/permissions/translations in an external array that we could just loop through for generating these lists of links/routes....

skomorokh: LOL like how the Settings component we're replacing worked?

skomorokh: I think Jeffrey intentionally didn't do that so that the routes would actually be in index.js as per the intention the whole routing for the module be laid out there.

skomorokh: ...so we can have our cake and eat it too: just export the routes from index. There's still a layer of indirection by splitting it out into an object but I agree that's better than the extent of duplication.

PR #303, opened 2019-04-04
Fix regex for the last item in each rule. Part of UICIRC-210

mkuklis:
@zburke started seeing issues with integration tests. It looks like the last item in each rule was not correctly converted to UUID. This PR is trying to address it which hopefully makes the tests happy again.

### DIFF
--- a/src/settings/CirculationRules.js
+++ b/src/settings/CirculationRules.js
@@ -181,8 +181,9 @@ class CirculationRules extends React.Component {
     const records = this.getRecords();
     return records.reduce((memo, r) => {
       // eslint-disable-next-line no-useless-escape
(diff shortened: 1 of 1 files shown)


marcjohnson-kint:
Does that mean that any editing of the rules would mean the last rule was invalid?

zburke:
If substitution fails for any part of a rule then that whole rule becomes invalid. This is correct behavior, IMHO, though we should warn if substitution fails. The issue addressed here is that a bad reg-ex caused an erroneous substitution failure for the last item on the last line.

Since our integration tests function by adding a new rule to the end and then testing the renewal count of that rule, they failed because items were loaned under the penultimate circulation rule (which happened to have liberal renewal limits), not the new rule (with strict limits).

PR #309, opened 2019-04-11
UICIRC-231, UICIRC-159: Add asterisk. Change labels.

OleksandrAntonenko1:


maximdidenkoepam:
Are you sure that combine two tasks in one PR is a proper way?

OleksandrAntonenko1:
There is not a lot of changes in both of them, also changes are located in the same module. And Jira handles multiple stories in one PR. So for the small stories, I think it's ok.

PR #318, opened 2019-04-23
UICIRC-232 - added validation for text editor

maximdidenkoepam:
# Purpose
To rework validation for patron notice template form

# Link
https://issues.folio.org/browse/UICIRC-232

# Screenshoot
Screen Shot 2019-04-23 at 14 09 43

### DIFF
--- a/src/settings/PatronNotices/PatronNoticeForm.js
+++ b/src/settings/PatronNotices/PatronNoticeForm.js
@@ -219,10 +219,6 @@ class PatronNoticeForm extends React.Component {
     );
   };
 
-  // Synchronous validation functions
-  requireName = value => (value ? undefined : <FormattedMessage id="ui-circulation.settings.patronNotices.errors.nameRequired" />);
-  requireCategory = value => (value ? undefined : <FormattedMessage id="ui-circulation.settings.patronNotices.errors.categoryRequired" />);
-
   render() {
     const { handleSubmit, initialValues = {} } = this.props;
     const category = initialValues && initialValues.category;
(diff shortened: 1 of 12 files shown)


src/settings/Validation/engine/handlers.js
```diff
@@ -11,3 +11,8 @@ export const isIntegerGreaterThanZero = (value) => isInteger(value) && value > 0
 export const isPositiveNumber = (value) => isInteger(value) && value >= 0;
 
 export const isInInterval = (min, max, value) => value >= min && value <= max;
+
+export const isNotEmptyEditor = (value = '') => {
+  const plainText = value.replace(/<\/?[^>]+(>|$)/g, '');
+  return !isEmpty(plainText);
```
OleksandrAntonenko1: Please add an empty line before the return.

src/settings/PatronNotices/PatronNoticeForm.js
```diff
@@ -256,12 +252,14 @@ class PatronNoticeForm extends React.Component {
                 data-test-patron-notice-template-name
               >
                 <Field
-                  required
-                  label={<FormattedMessage id="ui-circulation.settings.patronNotices.notice.name" />}
+                  label={
```
OleksandrAntonenko1: It might be a good idea to use "required" instead of adding an asterisk manually, what do you think about it?

maximdidenkoepam: No, this attribute will broke our form validation, and also all required fields in circulation module follows the same pattern.

PR #321, opened 2019-04-26
UICIRC-184 - Prohibit notice policy deletion

maximdidenkoepam:


src/settings/NoticePolicy/NoticePolicySettings.js
```diff
@@ -80,6 +98,12 @@ class NoticePolicySettings extends React.Component {
         entryLabel={formatMessage({ id: 'ui-circulation.settings.noticePolicy.entryLabel' })}
         validate={validateNoticePolicy}
         defaultEntry={NoticePolicy.defaultNoticePolicy()}
+        isEntryInUse={this.isPolicyInUse}
```
zburke: You need to depend on stripes-components v2.6.0, which won't be available until stripes v2.6.0 is released, to use these props. Please update the stripes dep to v2.6.0 in your {{package.json}} file. That version is in-progress on CI, so if you point your NPM `@folio:registry` at `npm-folioci` or clone {{stripes}} into your workspace, you should be all set.

mkuklis:
@maximdidenkoepam do you think we can merge this one? The reason I'm asking is because I need to do something similar and I just want to steal your implementation :)

maximdidenkoepam:
@mkuklis I am ready to merge, but @zburke requested changes and I don't see approve from him :)

mkuklis:
Ah @zburke do things look good now to you?

PR #330, opened 2019-05-10
Add support for attributes with  multiple values to circ editor. Part of UICIRC-255

mkuklis:
https://issues.folio.org/browse/UICIRC-255

### DIFF
--- a/src/settings/CirculationRules.js
+++ b/src/settings/CirculationRules.js
@@ -168,12 +168,12 @@ class CirculationRules extends React.Component {
     } = this.props.resources;
 
     return [
-      ...patronGroups.records.map(r => ({ name: kebabCase(r.group), id: r.id, prefix: 'g' })),
-      ...materialTypes.records.map(r => ({ name: kebabCase(r.name), id: r.id, prefix: 'm' })),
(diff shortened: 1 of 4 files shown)


mkuklis:
@zburke I honestly don't know all the rules for this syntax but I think you can find more details about them here:

https://github.com/folio-org/mod-circulation/blob/master/doc/circulationrules.md

I also remember there was a nice presentation done by Filip about them at some point. I wonder if @julianladisch could point us to it.

The regex is pretty complex and I honestly don't know why we are doing it on the front end. I assume something similar has to be happening on the backend when we need to extract the uuids. It started being simpler but it looks like the syntax is pretty complex and I think at some point we may need to switch to a real tokenizer / lexer to handle it.

I apologize for not describing what has been fixed under this PR. I assumed JIRA had that info but here is a bit more details.

The previous implementation didn't cover the case where we used multiple values for a given type for example this worked fine:

`m book: l example-loan-policy r allow-all n alternate-notice-policy` 

but this would fail  (the `dvd` would not get converted to uuid):

`m book dvd: l example-loan-policy r allow-all n alternate-notice-policy`

zburke:
No apology necessary. I'm just thinkin' this stuff looks complicated and neither the code nor the ticket make any reference to the mod-circulation documentation that explains what's going on. The circulationrules.md file you linked to is a great reference.

PR #395, opened 2019-08-12
UICIRC-308 - Verify coding of "required" field indicator

maximdidenkoepam:
# Purpose
Get rid of asterisks in translation files

# Link
https://issues.folio.org/browse/UICIRC-308

# Approach
For both cases we can't use required attribute because FormattedMessage component is not a value of label property of Field component. In order to handle the issue new component was created. This component may be reused in another PR to handle similar issue for patron notice policies. 

# Screenshots
Screen Shot 2019-08-12 at 17 51 00
Screen Shot 2019-08-12 at 17 51 12

### DIFF
--- a/src/settings/RequestCancellationReasons.js
+++ b/src/settings/RequestCancellationReasons.js
@@ -4,7 +4,7 @@ import { FormattedMessage } from 'react-intl';
 
 import { ControlledVocab } from '@folio/stripes/smart-components';
 import { stripesConnect } from '@folio/stripes/core';
-
+import { Asterisk } from './components';
 
 class RequestCancellationReasons extends React.Component {
   static propTypes = {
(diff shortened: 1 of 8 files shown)


src/settings/RequestCancellationReasons.js
```diff
@@ -30,7 +30,12 @@ class RequestCancellationReasons extends React.Component {
         visibleFields={['name', 'description', 'publicDescription']}
         hiddenFields={['lastUpdated', 'numberOfObjects']}
         columnMapping={{
-          name: <FormattedMessage id="ui-circulation.settings.cancelReasons.labelShort" />,
+          name: (
+            <React.Fragment>
+              <FormattedMessage id="ui-circulation.settings.cancelReasons.labelShort" />
+              <Asterisk />
+            </React.Fragment>
```
dltj: I don't think this is correct.  What is needed is a `required` attribute on the Field.  See https://github.com/folio-org/ui-inventory/blob/master/src/edit/InstanceForm.js#L484 for an example.

maximdidenkoepam: It's not possible, because in both cases FormattedMessage component which renders label does not belong to any Field component as label property.

dltj: I see what you mean.
image

I'm not entirely comfortable with how this implementation is different from how the `required` attribute is used on `<Field>` tags.  I'm going to take this into the "#stripes" channel on Slack for advice.

maximdidenkoepam:
@dltj, @zburke done.

src/settings/RequestCancellationReasons.js
```diff
@@ -30,7 +32,14 @@ class RequestCancellationReasons extends React.Component {
         visibleFields={['name', 'description', 'publicDescription']}
         hiddenFields={['lastUpdated', 'numberOfObjects']}
         columnMapping={{
-          name: <FormattedMessage id="ui-circulation.settings.cancelReasons.labelShort" />,
+          name: (
+            <Label
+              className={css.header}
```
zburke: I don't think this is necessary any longer now that 
[STCOM-565](https://issues.folio.org/browse/STCOM-565) / folio-org/stripes-components/pull/1025 is in place. Can we remove it?

maximdidenkoepam: Done

PR #404, opened 2019-08-20
console cleanup

zburke:
I don't remember what sent me off on this quest, but I discovered these
not-quite-right problems with props, throwing warnings in the console,
along the way. Now, the console is clean.


src/settings/StaffSlips/StaffSlipManager.js
```diff
@@ -53,26 +55,29 @@ class StaffSlipManager extends React.Component {
...
-        entryFormComponent={StaffSlipForm}
-        validate={this.validate}
-        nameKey="name"
-        permissions={{
-          put: 'ui-circulation.settings.staff-slips',
-          post: 'ui-circulation.settings.staff-slips.post',
-          delete: 'ui-circulation.settings.staff-slips.delete'
-        }}
-      />
+      <IntlConsumer>
```
aditya-matukumalli: Could we just avoid _IntlConsumer_ here altogether by utilizing the render props of _FormattedMessage_. Similar to what we have [here](https://github.com/folio-org/ui-users/blob/master/src/settings/permissions/PermissionSets.js#L65)

zburke: Technically yes, and I see the appeal of that simplicity. I probably went for the `<IntlConsumer>` approach out of habit from other situations where there were multiple values requiring translation to strings, so getting `intl` is helpful. So ... I dunno. Render props here would be simpler; `<IntlConsumer>` here would be more consistent. 🤷‍♀

aditya-matukumalli: Ah okay. Either approach is fine I think. If its more consistent across the app, lets go with the current implementation.

zburke: Eh, I went for your approach favoring simplicity :)

PR #406, opened 2019-08-22
UICIRC-285: Add backward moving feature to organization hierarchy menus in circulation rules editor

VictorSoroka1:
# Purpose
Implement moving to the previous section in current sections list upon click on Backspace button. Implement hint selecting functionality upon mouse click. Refs [UICIRC-285](https://issues.folio.org/browse/UICIRC-285).

# Screenshots

- clicking on Backspace button
backspace

- clicking on a mouse button
mouse

### DIFF
--- a/src/settings/lib/RuleEditor/rules-show-hint.js
+++ b/src/settings/lib/RuleEditor/rules-show-hint.js
@@ -16,6 +16,9 @@ import {
   HINT_ELEMENT_CLASS,
 } from '../../../constants';
 
+const HINT_SECTIONS_CONTAINER = 'CodeMirror-hints-sections-container';
+const HINT_SECTION_CONTAINER = 'CodeMirror-hints-list';
+
 CodeMirror.registerHelper('hint', 'auto', { resolve: resolveAutoHints });
 
 const defaultOptions = {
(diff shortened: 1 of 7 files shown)


src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -322,43 +323,70 @@ class Widget {
...
 
-        if (completion.options.completeOnSingleClick) {
-          this.pick();
-        }
-      }
+  handleSectionHintClick(targetElement) {
+    const hint = Widget.getHintElement(targetElement);
+    const hintExists = hint && hint.hintId != null;
+
+    if (!hintExists) return;
```
SergiyVSergiyenko: Since the 'hintExists' variable is no longer used here, I would prefer not to create it at all.
```if (!hint || hint.hintId == null) return;```

VictorSoroka1: I would prefer to leave as is as having more descriptive variables better than trying to find the place where it should be used. I see the point to not create variables when you can utilize `isEmpty` from `lodash` which makes it clear what is meant. Or when there is just check like `if (item)`.

PR #413, opened 2019-09-04
UICIRC-304 - Change "Edit patron notice template page" to support multiple-item templates

maximdidenkoepam:
# Purpose
Add ability use loops for tokens in template editor.

# Link
https://issues.folio.org/browse/UICIRC-304

# Screenshots
Screenshot_5
Screenshot_6

### DIFF
--- a/src/settings/components/TemplateEditor/TemplateEditor.js
+++ b/src/settings/components/TemplateEditor/TemplateEditor.js
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import ReactQuill, { Quill } from 'react-quill';
 
 import {
-  isEmpty,
   isNull,
   forEach,
+  includes,
 } from 'lodash';
 
 import {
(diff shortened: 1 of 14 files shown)


src/settings/utils/template-resolver.js
```diff
@@ -2,7 +2,7 @@ export default function (str) {
   return (o) => {
     return str.replace(/{{([^{}]*)}}/g, (a, b) => {
       const r = o[b];
-      return typeof r === 'string' || typeof r === 'number' ? r : a;
+      return typeof r === 'string' || typeof r === 'number' ? r : '';
```
VictorSoroka1: Since you edited stuff here, could you please give more descriptive variable names to `o`, `r`, and maybe `a` and `b` as it is hard to reason about them.

src/settings/components/TemplateEditor/TemplateEditor.js
```diff
@@ -128,6 +134,32 @@ class TemplateEditor extends React.Component {
...
+  insertRepeatingTokens = (tokens, tag) => {
+    let tags = '';
+    const startTag = `{{#${tag}}}`;
+    const endTag = `{{/${tag}}}`;
+
+    const editor = this.quill.current.getEditor();
+    const editorContent = editor.getText();
+    const loopExists = includes(editorContent, startTag) && includes(editorContent, endTag);
+
+    if (loopExists) {
```
VictorSoroka1: I would suggest simplifying things there a little bit

```
let tags = map(tokens, (token) => token); or just Object.values(tokens)

if (!loopExists) {
  tags = [startTag, ...tokens, endTag];
}

tags = tags.join('\n');
```
Does it make sense?

maximdidenkoepam: No, take a look on if statement. Depending on condition we should add token with '\n' on different positions. Also in case if loop exists we should add end tag with '\n'. If i will follow your approach the code will look like:
```
tags = loopExists 
  ? tags = map(tokens, token => `{{${token}}}\n`)
  : [startTag, map(tokens, token => `\n{{${token}}}`), '\n', endTag];

editor.insertText(cursorPosition, tags.join(''));
```
Not sure that this code is more readable.
Moreover in this case we receive strange behavior for the cursor position.

VictorSoroka1: Agree.

PR #414, opened 2019-09-05
Fix accordion tests

maximdidenkoepam:
# Purpose
Fix tests broken by updates in stripes-components


zburke:
Thank you, @maximdidenkoepam, for fixing all these tests!

PR #421, opened 2019-09-13
UICIRC-316, UICIRC-319, UICIRC-320: Add delivery tokens, delete active checkbox on settings

Yurii-Danylenko:
https://issues.folio.org/browse/UICIRC-316
https://issues.folio.org/browse/UICIRC-319
https://issues.folio.org/browse/UICIRC-320

## Purpose
Add a set of tokens that will be available for delivery slips. Remove a couple of redundant tokens. Delete "active" checkbox from all staff slip edit pages as non-functional. Change save button title from "Save and Close" to "Save". Replace current button translations with stripes-core


src/settings/StaffSlips/StaffSlipForm.js
```diff
@@ -68,7 +67,7 @@ class StaffSlipForm extends React.Component {
     const { pristine, submitting, initialValues } = this.props;
     const edit = initialValues && initialValues.id;
     const saveLabel = edit
-      ? <FormattedMessage id="ui-circulation.settings.staffSlips.saveAndClose" />
+      ? <FormattedMessage id="stripes-core.button.save" />
```
maximdidenkoepam: Why do we need to change the label?

Yurii-Danylenko: It is a requirement of the story https://issues.folio.org/browse/UICIRC-316. 4-th point

PR #425, opened 2019-09-17
UICIRC-314: Add location items filtering into the circulation rules editor menu

YevheniiMaltsev:
## Purpose

Add location items filtering into the circulation rules editor menu in the scope of the [story](https://issues.folio.org/browse/UICIRC-314).

## Screenshots

filtering_1
filtering_2
filtering_3
filtering_4

### DIFF
--- a/src/settings/lib/RuleEditor/rules-show-hint.js
+++ b/src/settings/lib/RuleEditor/rules-show-hint.js
@@ -309,8 +309,18 @@ class Widget {
     if (completion.options.closeOnUnfocus) {
       let closingOnBlur;
 
-      cm.on('blur', this.handleBlur = () => { closingOnBlur = setTimeout(completion.close, 100); });
+      cm.on('blur', this.handleBlur = (cmInstance, e) => {
+        const { relatedTarget } = e;
+
+        if (!relatedTarget || !isChildTextInputField(this.container, relatedTarget)) {
+          closingOnBlur = setTimeout(completion.close, 100);
+        }
+      });
       cm.on('focus', this.handleFocus = () => { clearTimeout(closingOnBlur); });
+
+      CodeMirror.on(cm, 'blurHintInput', this.handleInputBlur = () => {
+        closingOnBlur = setTimeout(completion.close, 100);
+      });
     }
 
     const startScroll = cm.getScrollInfo();
@@ -337,8 +347,11 @@ class Widget {
       this.handleSectionHintClick(e.target);
     });
 
-    CodeMirror.on(this.container, 'mousedown', () => setTimeout(() => cm.focus(), 20));
-
+    CodeMirror.on(this.container, 'mousedown', (e) => {
+      if (e.target && !isChildTextInputField(this.container, e.target)) {
+        setTimeout(() => cm.focus(), 20);
+      }
+    });
     CodeMirror.signal(data, 'select', completions[0], this.getCurrentSection().getListNodeAt(0));
   }
 
@@ -516,6 +529,7 @@ class Widget {
     if (closeOnUnfocus) {
       cm.off('blur', this.handleBlur);
       cm.off('focus', this.handleFocus);
+      CodeMirror.off(cm, 'blurHintInput', this.handleInputBlur);
     }
 
     cm.off('scroll', this.handleScroll);
@@ -600,7 +614,7 @@ class HintSection {
     this.itemsOptions = sectionOptions.list;
 
     if (sectionOptions.header) {
-      this.container.appendChild(createHeader(sectionOptions.header, 'CodeMirror-hints-subheader'));
+      this.initHeader(sectionOptions.header);
     }
 
     this.listContainer = document.createElement('ul');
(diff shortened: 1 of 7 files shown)


src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -309,7 +309,13 @@ class Widget {
     if (completion.options.closeOnUnfocus) {
       let closingOnBlur;
 
-      cm.on('blur', this.handleBlur = () => { closingOnBlur = setTimeout(completion.close, 100); });
+      cm.on('blur', this.handleBlur = (cmInstance, e) => {
+        const { relatedTarget } = e;
+
+        if (!relatedTarget || (relatedTarget && !isTextInputField(relatedTarget))) {
```
VictorSoroka1: Could you please add more context for this check. It can be just several new variables.

src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -782,11 +851,33 @@ class MultipleSelectionHintSection extends HintSection {
...
   }
+
+  updateSectionWidth() {
+    const headerStyle = getComputedStyle(this.header);
+
+    this.listContainer.style.width = headerStyle.width;
+    this.listContainer.style.margin = headerStyle.margin;
+  }
+
+  getListNodesArray() {
```
VictorSoroka1: Since here no function params are not needed I suggest making it as [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get).
I also suggest not mentioning the structure name as `Array` in this case. We want to be abstract on such things.
```suggestion
  get listNodes() {
```
so in code later it can be used without braces (e.g. `this.listNodes.length`).

src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -698,6 +713,10 @@ class HintSection {
     this.selectedHintIndex = index;
   }
 
+  getListNodesArray() {
```
VictorSoroka1: Please convert it to getter `listNodes`.

src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -782,11 +851,33 @@ class MultipleSelectionHintSection extends HintSection {
...
+
+  getListNodesArray() {
+    return this.getDisplayedNodes();
+  }
+
+  getListNodeAt(index) {
+    return this.getDisplayedNodes()[index];
+  }
+
+  getDisplayedNodes() {
```
VictorSoroka1: Please convert it to a getter.

src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -709,6 +728,56 @@ class MultipleSelectionHintSection extends HintSection {
...
+  filterItems(filterValue) {
+    if (isEmpty(this.itemsOptions)) return;
+
+    const filteredItems = this.itemsOptions.filter(itemOptions => {
+      return itemOptions.displayText.toLowerCase().includes(filterValue.toLowerCase());
+    });
+
+    const itemsIdsToDisplay = filteredItems.map(itemOptions => itemOptions.id);
+
+    super.getListNodesArray().forEach((currentListItem) => {
```
VictorSoroka1: for consistency with a `listNodes` naming
```suggestion
    super.getListNodesArray().forEach((listNode) => {
```

src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -709,6 +728,56 @@ class MultipleSelectionHintSection extends HintSection {
...
+      listNode.classList.toggle(ACTIVE_HINT_ELEMENT_CLASS, false);
+    });
+  }
+
+  filterItems(filterValue) {
+    if (isEmpty(this.itemsOptions)) return;
+
+    const filteredItems = this.itemsOptions.filter(itemOptions => {
+      return itemOptions.displayText.toLowerCase().includes(filterValue.toLowerCase());
+    });
```
VictorSoroka1: I suggest to get rid of the `itemsIdsToDisplay` variable and add that map right after the filter here. What do you think about it?
```suggestion
    }).map(itemOptions => itemOptions.id);
```

src/settings/lib/RuleEditor/rules-show-hint.js
```diff
@@ -709,6 +728,56 @@ class MultipleSelectionHintSection extends HintSection {
...
     this.initCompletionButton(sectionOptions.buttonText);
+
+    if (!this.header) {
+      this.initHeader('');
+    }
+
+    this.initFilterField(sectionOptions.filterPlaceholder);
+  }
+
+  initFilterField(filterPlaceholder) {
```
VictorSoroka1: I like the idea that dependent methods like this are defined after the place where it is called as you did. Just wanted to emphasize it. 👍

src/settings/lib/RuleEditor/rules-show-hint.js
YevheniiMaltsev: I think, after suggested changes the code will look more complicated. So I would like to leave this changes as it is. Are you okey with it?

src/settings/lib/RuleEditor/rules-show-hint.js
YevheniiMaltsev: Thanks :)

src/settings/lib/RuleEditor/rules-show-hint.js
VictorSoroka1: Sounds good.

PR #433, opened 2019-09-20
UICIRC-296: Fix incorrect save button behavior in Circulation Rules E…

SergiyVSergiyenko:
### Purpose
Fix incorrect save button behavior in Circulation Rules Editor according to [story](https://issues.folio.org/browse/UICIRC-296).

**Demo**

correct_save_button_behavior

### DIFF
--- a/src/settings/CirculationRules.js
+++ b/src/settings/CirculationRules.js
@@ -187,7 +187,9 @@ class CirculationRules extends React.Component {
 
   onSubmit(values) {
     const rules = values.rules.replace(/\t/g, '    ');
+
     this.save(rules);
+    this.setState({ rules: values.rules });
   }
 
   getRules() {
@@ -351,7 +353,7 @@ class CirculationRules extends React.Component {
       return (<div />);
     }
 
-    const rules = this.getRules();
+    const rules = this.state.rules || this.getRules();
     const editorProps = this.getEditorProps();
 
     return (
--- a/test/bigtest/interactors/circulation-rules.js
+++ b/test/bigtest/interactors/circulation-rules.js
@@ -108,12 +108,17 @@ const scrollingOffset = 3;
 @interactor class Editor {
   static defaultScope = '.react-codemirror2';
 
-  setValue = action(function (value, showHint = true) {
+  setValue = action(function (value, overrideOptions) {
+    const defaultOptions = { showHint: true, append: false };
+    const options = Object.assign(defaultOptions, overrideOptions);
+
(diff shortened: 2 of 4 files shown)


src/settings/CirculationRules.js
```diff
@@ -187,7 +187,9 @@ class CirculationRules extends React.Component {
 
   onSubmit(values) {
     const rules = values.rules.replace(/\t/g, '    ');
+
     this.save(rules);
+    this.setState({ ...values });
```
VictorSoroka1:
```suggestion
    this.setState({ rules });
```
Should it be like that? As for now, you are saving all properties to the state.

test/bigtest/interactors/circulation-rules.js
```diff
@@ -121,6 +121,20 @@ const scrollingOffset = 3;
     }).run();
   });
 
+  changeValue = action(function (value, showHint = true) {
```
YevheniiMaltsev: I'm not sure that we need to create a new method for it. At least, we can try to reuse a setValue because there is a very similar body for now.

@SergiyVSergiyenko what do you think about it?

VictorSoroka1: I would get rid of that method and extended `setValue` one.

test/bigtest/tests/circulation-rules-test.js
```diff
@@ -1054,5 +1054,37 @@ describe('CirculationRules', () => {
...
+      it('should enable save button', () => {
+        expect(circulationRules.isSaveButtonDisabled).to.be.false;
+      });
+
+      describe('clicking save button', () => {
+        beforeEach(async function () {
+          await circulationRules.clickSaveRulesBtn();
+        });
+
+        it('should have disabled save button', () => {
```
VictorSoroka1:
```suggestion
        it('should disable save button upon succesfull save', () => {
```

test/bigtest/tests/circulation-rules-test.js
```diff
@@ -1054,5 +1054,37 @@ describe('CirculationRules', () => {
     it('should choose loan policy as a fallback', () => {
       expect(savedRules).to.equal(`m 1a54b431-2e4f-452d-9cae-9cee66c9a892 5ee11d91-f7e8-481d-b079-65d708582ccc: l ${lPolicy.id} r ${rPolicy.id} n ${nPolicy.id}`);
     });
+
```
YevheniiMaltsev: It looks like this test case is very similar to the previous one in this file. I seems, they can be combined

PR #439, opened 2019-09-27
UICIRC-338 Save validation error on Fixed Due Date Schedule Form

JohnC-80:
[UICIRC-338](https://issues.folio.org/browse/UICIRC-338)

Fixes:
* Form saves as expected with multiple date range lines
* New line date schedule line no longer appears with `invalid date` messages
* Implemented a `<PaneFooter>`


JohnC-80:
@zburke changing `unshift` to `push` was the lowest impact at App level (seems odd that you'd have date ranges listed in reverse order/input them in reverse order but maybe that's what's necessary and I'm just flat out of context). 
At the STCOM level, Datepickers need work or some extra app-level attention to function normally for use-cases like this (lists of editable dates.) They'll present values, but that doesn't accurately represent the state since communication is implemented one-way / value prop is initialization-only with the current implementation. This can be circumvented by updating the `key` prop on the Datepicker, but yeah - usual behavior is definitely desired and would make a few of our lives easier.
Filed STCOM-606 for this.

So with the old way, the list would unshift with an empty value, but the pickers wouldn't be updated to display the empty state - so the data was actually empty when it appeared as filled in. Of course, empty data => failed validation. The last row would initialize with the unshift, but since `dateFormat` wasn't set on the pickers, `invalid date` would show up.

A weird gotcha with editable lists like this one - the actual rendering order - usually performed by a `map()` - wouldn't accurately reflect the order of the data/changes in order of the data - part of reason why React generally says that index-based keys are a bad idea.

PR #441, opened 2019-09-30
UICIRC-333: Add full hierarchical path to location codes in the circulation rules editor field

YevheniiMaltsev:
## Purpose

Add full hierarchical path to location codes in the circulation rules editor field in the scope of the [story](https://issues.folio.org/browse/UICIRC-333).

## Screenshots

full_path_1
full_path_2
full_path_3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Validate for closed loans and fines/fees in loan history settings. Refs UITEN-42.
 * Fix incorrect save button behavior in Circulation Rules Editor. Refs UICIRC-296.
 * Add handling of the special item Any for intermediate menus. Refs UICIRC-334.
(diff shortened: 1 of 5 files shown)


src/settings/utils/code-formatters.js
```diff
@@ -0,0 +1,37 @@
+import { find } from 'lodash';
```
VictorSoroka1: Good idea moving all these to utils. I wonder about prepend the file name with location to make it `location-code-formatters` in order to add more context. What do you think about that?

YevheniiMaltsev: @VictorSoroka1 thanks for you comment. I renamed the file.

YevheniiMaltsev:
@mkuklis thanks :)

PR #442, opened 2019-10-03
UITEN-57: Clean up Loan history model before sending to backend

SergiyVSergiyenko:
### Purpose
Clean up Loan history model before sending to the backend according to the [story](https://issues.folio.org/browse/UITEN-57).

**Demo**

Scenario 3

Scenario 4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Add handling of the special item Any for intermediate menus. Refs UICIRC-334.
 * Add full hierarchial path to location codes that are used in the input field of the circulation rules editor. Refs UICIRC-333.
 * Fix loan policy holds section. Refs UICIRC-349.
+* Clean up Loan history model. Refs UITEN-57.
 
 ## [1.11.0](https://github.com/folio-org/ui-circulation/tree/v1.13.0) (2019-09-13)
 [Full Changelog](https://github.com/folio-org/ui-circulation/compare/v1.10.0...v1.11.0)
--- a/src/settings/LoanHistory/LoanHistoryForm.js
+++ b/src/settings/LoanHistory/LoanHistoryForm.js
@@ -8,6 +8,8 @@ import {
   Field,
   FieldArray,
   getFormValues,
+  arrayRemoveAll,
+  change,
 } from 'redux-form';
 
 import stripesForm from '@folio/stripes/form';
@@ -27,11 +29,13 @@ import {
   closingTypes,
   closedLoansRules,
 } from '../../constants';
+import { normalize } from './utils/normalize';
 
 import css from './LoanHistoryForm.css';
 
 class LoanHistoryForm extends Component {
   static propTypes = {
+    dispatch: PropTypes.func.isRequired,
     handleSubmit: PropTypes.func.isRequired,
     onSubmit: PropTypes.func.isRequired,
     pristine: PropTypes.bool,
@@ -47,7 +51,16 @@ class LoanHistoryForm extends Component {
     this.state = { checked: false };
   }
 
-  onSave = data => this.props.onSubmit({ loan_history: JSON.stringify(data) });
+  onSave = data => {
+    const {
+      dispatch,
+      onSubmit,
+    } = this.props;
+
+    const normalizedData = normalize({ data, dispatch });
+
+    onSubmit({ loan_history: JSON.stringify(normalizedData) });
+  };
 
   getLastMenu = () => {
     const {
@@ -67,7 +80,16 @@ class LoanHistoryForm extends Component {
     );
   }
 
+  // Due to the issue https://github.com/erikras/redux-form/issues/4101
+  // the multiple actions should be dispatched instead of single one ('clearFields')
   toggleCheckbox = () => {
+    const { dispatch } = this.props;
+
+    dispatch(change('loanHistoryForm', 'closingType.feeFine', null));
+    dispatch(change('loanHistoryForm', 'closingType.loanExceptions', []));
+    dispatch(change('loanHistoryForm', 'feeFine', {}));
+    dispatch(arrayRemoveAll('loanHistoryForm', 'loanExceptions'));
+
     this.setState(({ checked }) => ({
       // eslint-disable-next-line react/no-unused-state
       checked: !checked
--- a/src/settings/LoanHistory/utils/normalize.js
+++ b/src/settings/LoanHistory/utils/normalize.js
@@ -1,0 +1,43 @@
+import {
+  cloneDeep,
+  forEach,
+  set,
+  isArray,
+  isString,
+} from 'lodash';
+import { change } from 'redux-form';
+
+import { closingTypesMap } from '../../../constants';
+
+const cleanInterval = ({ data, dispatch }) => {
+  const newData = cloneDeep(data);
+
+  forEach(data.closingType, (value, key) => {
+    if (isString(value) && value !== closingTypesMap.INTERVAL) {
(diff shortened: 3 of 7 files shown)


test/bigtest/tests/loan-history/loan-history-form-test.js
```diff
@@ -121,18 +121,49 @@ describe('Loan History Form', () => {
...
+    describe('selecting any radio button', () => {
       beforeEach(async () => {
         await LoanHistoryForm
           .feeFine.selectRadio(closingTypesMap.IMMEDIATELY)
           .feeFine.selectRadio(closingTypesMap.INTERVAL)
           .feeFine.selectRadio(closingTypesMap.NEVER);
+      });
 
-        await LoanHistoryForm.saveButton.click();
+      describe('when checkbox is unchecked', () => {
```
YevheniiMaltsev:
```suggestion
      describe('unchecking the checkbox', () => {
```

src/settings/LoanHistory/utils/normalize.js
```diff
@@ -0,0 +1,48 @@
...
+  forEach(callbacks, callback => {
+    filteredEntity = callback(filteredEntity);
+  });
+
+  return filteredEntity;
+};
+
+
+export const normalize = entity => {
+  const callbacks = [checkInterval];
```
YevheniiMaltsev: I'm not sure, that callbacks is a right word to describe the array purposes. Maybe, it should be renamed to "handlers", for example.

src/settings/LoanHistory/utils/normalize.js
```diff
@@ -0,0 +1,48 @@
...
+  isString,
+} from 'lodash';
+import { change } from 'redux-form';
+
+import { closingTypesMap } from '../../../constants';
+
+const checkInterval = ({ data, dispatch }) => {
+  const newData = cloneDeep(data);
+
+  forEach(Object.entries(data.closingType), ([key, value]) => {
```
OleksandrYehorovEPAM:
```suggestion
  forEach(data.closingType, (key, value) => {
```
Should do the same thing

src/settings/LoanHistory/utils/normalize.js
```diff
@@ -0,0 +1,48 @@
...
+          set(newData, [key, index], {});
+          dispatch(change('loanHistoryForm', [key, index], {}));
+        }
+      });
+    }
+  });
+  return newData;
+};
+
+const filter = (entity, ...callbacks) => {
```
YevheniiMaltsev: I'm not sure that we need to have this function here because its name is a little bit confusing and we can try to do this job in scope of the normalize function.

@VictorSoroka1 What do you think about it?

test/bigtest/tests/loan-history/loan-history-form-test.js
```diff
@@ -217,18 +248,37 @@ describe('Loan History Form', () => {
...
-        it('should save form successfully', () => {
-          expect(LoanHistoryForm.callout.successCalloutIsPresent).to.be.true;
+        describe('when checkbox is unchecked', () => {
+          beforeEach(async () => {
+            await LoanHistoryForm.clickTreatEnabledCheckbox();
+          });
+
+          it('should disappear section for closed loans with associated fees/fines', () => {
+            expect(LoanHistoryForm.feefineSection.isPresent).to.be.false;
+          });
```
YevheniiMaltsev: It seems, a blank line is missed

test/bigtest/tests/loan-history/loan-history-form-test.js
```diff
@@ -121,18 +121,49 @@ describe('Loan History Form', () => {
...
+            await LoanHistoryForm.clickTreatEnabledCheckbox();
+          });
+
+          const check = items => {
+            return items.some(item => item.isChecked);
+          };
+
+          it('should display section for closed loans with associated fees/fines', () => {
+            expect(LoanHistoryForm.feefineSection.isPresent).to.be.true;
+          });
```
YevheniiMaltsev: It looks like a blank line is missed

test/bigtest/tests/loan-history/loan-history-form-test.js
```diff
@@ -121,18 +121,49 @@ describe('Loan History Form', () => {
...
+        it('should disappear section for closed loans with associated fees/fines', () => {
+          expect(LoanHistoryForm.feefineSection.isPresent).to.be.false;
+        });
+
+        describe('when checkbox is checked', () => {
+          beforeEach(async () => {
+            await LoanHistoryForm.clickTreatEnabledCheckbox();
+          });
+
+          const check = items => {
```
YevheniiMaltsev: Could you please clarify if there are any reasons to declare a global function and use it only in one place?

I believe, we can try just to declare a constant isChecked and assign a result to it (Line 150-151), or add this util function to the interactor.

@SergiyVSergiyenko What do you think about it?

src/settings/LoanHistory/utils/normalize.js
SergiyVSergiyenko: More exactly
```
forEach(data.closingType, (value, key) => {
```
Anyway thanks a lot for the hint.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * Fix incorrect save button behavior in Circulation Rules Editor. Refs UICIRC-296.
 * Add handling of the special item Any for intermediate menus. Refs UICIRC-334.
 * Add full hierarchial path to location codes that are used in the input field of the circulation rules editor. Refs UICIRC-333.
+* Clean up Loan history model before sending to backend. Refs UITEN-57.
```
YevheniiMaltsev: It seems, some date will be cleaned even without sending to backend. Am I right?

src/settings/LoanHistory/LoanHistoryForm.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-unused-state */
```
VictorSoroka1: I am more inclined to have eslint-disable comments in place rather than disabling the whole file.

src/settings/LoanHistory/LoanHistoryForm.js
```diff
@@ -43,11 +48,18 @@ class LoanHistoryForm extends Component {
...
     this.state = { checked: false };
   }
 
-  onSave = data => this.props.onSubmit({ loan_history: JSON.stringify(data) });
+  onSave = data => {
+    const {
+      dispatch,
+      onSubmit
+    } = this.props;
+    const newData = normalize({ data, dispatch });
```
VictorSoroka1: Please add an empty line after the props/state declarations.

src/settings/LoanHistory/LoanHistoryForm.js
```diff
@@ -67,9 +79,17 @@ class LoanHistoryForm extends Component {
     );
   }
 
+  // Due to the issue https://github.com/erikras/redux-form/issues/4101, has to use
+  // multiple dispatches of actions instead of single dispatch of 'clearFields' action.
```
VictorSoroka1:
```
// Due to the issue https://github.com/erikras/redux-form/issues/4101,
// multiple dispatches of actions should be used instead of single dispatch of 'clearFields' action.
```

src/settings/LoanHistory/utils/normalize.js
VictorSoroka1: Agree, on top of that in the current approach I do not like that we have two calls of `cloneDeep` function. I would rather not go for it if possible.

PR #444, opened 2019-10-03
Fix loan policy holds section. Refs UICIRC-349

mkuklis:
## Purpose
The PR makes the loan policy holds section be open by default. It also fixes some minor issues related to `Alternate loan period at renewal for items with an active, pending hold request` field visibility.

## Link
https://issues.folio.org/browse/UICIRC-349

### DIFF
--- a/src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.js
+++ b/src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import get from 'lodash/get';
+import { get } from 'lodash';
 import PropTypes from 'prop-types';
 import { Field } from 'redux-form';
 import {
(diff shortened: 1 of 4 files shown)


src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import get from 'lodash/get';
```
zburke: I think the preferred syntax is 
```
import { get } from 'lodash';
```
@ryandberger is that correct? I recall about preferring this, maybe due to treeshaking. Seems like this should be documented in stripes.

mkuklis: Sounds good @zburke. I just changed it.

ryandberger: @zburke, since we're using [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash), the output automatically transpiles to what @mkuklis had before anyway. https://www.azavea.com/blog/2019/03/07/lessons-on-tree-shaking-lodash/

PR #477, opened 2019-11-20
Update CirculationRules render

mkuklis:
Refs https://github.com/folio-org/ui-circulation/pull/474

### DIFF
--- a/src/settings/CirculationRules.js
+++ b/src/settings/CirculationRules.js
@@ -1,4 +1,7 @@
-import { kebabCase } from 'lodash';
+import {
+  kebabCase,
+  get,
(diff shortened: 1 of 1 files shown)


zburke:
Replaces #477, potentially. We only needed #477 because stripes-connect wasn't doing its job in the way most folks probably anticipated, i.e. even though these resources were specified with `resourceShouldRefresh: true`, the resource _wouldn't_ always refresh, or at least it would not indicate that it was doing so, which led to general unreliability.

VictorSoroka1:
Surprisingly tests started to pass in ui-circulation. This is great news.

PR #478, opened 2019-11-21
UICIRC-390: add item limit field to loan policy

annamelnyk:
### Description
Add new field Item Limit to Loan Policy
### Issue
https://issues.folio.org/browse/UICIRC-390
### Screenshot
image


src/settings/Validation/engine/handlers.js
```diff
@@ -16,6 +16,8 @@ export const isPositiveNumber = (value) => isInteger(value) && value >= 0;
 
 export const isInInterval = (min, max, value) => value >= min && value <= max;
 
+export const isInIntervalOrNull = (min, max, value) => !value || (value >= min && value <= max);
```
VictorSoroka1:
```suggestion
export const isInIntervalOrNull = (min, max, value) => !value || isInInterval(min, max, value);
```

src/settings/Validation/engine/validators.js
```diff
@@ -48,4 +49,8 @@ export default {
     validate: isOverdueFindGreaterThan,
     message: <FormattedMessage id="ui-circulation.settings.validate.greaterThanZero" />,
   },
+  isItemLimitIsInInterval: {
+    validate: isInIntervalOrNull.bind(null, 1, 9999),
```
VictorSoroka1:
```suggestion
    validate: value => isInIntervalOrNull(1, 9999, value),
```
Arrow functions like `isInIntervalOrNull` does not have context, so using `bind` of them can be misleading. I would rather simplify the mental work with the suggested code above.

src/settings/LoanPolicy/components/EditSections/LoansSection/LoansSection.js
```diff
@@ -150,14 +153,28 @@ class LoansSection extends React.Component {
...
         { policy.isLoanable() &&
-          <div data-test-loans-section-grace-period>
-            <Period
-              fieldLabel="ui-circulation.settings.loanPolicy.gracePeriod"
-              inputValuePath="loansPolicy.gracePeriod.duration"
-              selectValuePath="loansPolicy.gracePeriod.intervalId"
-              intervalPeriods={this.generateOptions(intervalPeriods, 'ui-circulation.settings.loanPolicy.selectInterval')}
-              changeFormValue={change}
-            />
+          <div>
```
VictorSoroka1:
```suggestion
          <Fragment>
```
Will fragment help to avoid wrapper div? Make sure to import Fragment from React if it will.

PR #479, opened 2019-11-21
UIU-1315: Add missing words "late" and "Set cost" on entry of new policy

arturorc:
## Description
Add missing words "late" and "Set cost" on entry of new policy to fix [UIU-1315](https://issues.folio.org/browse/UIU-1315)

Screenshot_2019-11-21 Settings - FOLIO


arturorc:
> When adding a new string, add it to `en.json` only. This change will be picked by Lokalise, added to the other locale files, and fed back into this repo in a separate PR, overwriting whatever changes are made here.

Ok updated

PR #493, opened 2019-12-05
UICIRC-400 - Provide two options for barcode tokens on patron notices

maximdidenkoepam:
# Purpose
- add tokens for item and user barcode in order to show barcode image
- allow old user and item barcode tokens show barcode as a plain string instead of image

# Link
https://issues.folio.org/browse/UICIRC-400

# Dependency
https://issues.folio.org/browse/MODTEMPENG-34

# Screenshots
Screen Shot 2019-12-05 at 11 58 04
Screen Shot 2019-12-05 at 11 58 18
Screen Shot 2019-12-05 at 11 58 45


VictorSoroka1:
Please add entry to the Changelog.

PR #496, opened 2019-12-06
UICIRC-405: Remove periods values from default config in `<LoanHistory>`, handle CSS issue when the loan history form is being submitted and add tests for anonymizing intervals.

SergiyVSergiyenko:
## Purpose
Remove periods values from default config in `<LoanHistory>`, handle CSS issue when the loan history form is being submitted and add tests for anonymizing intervals. [Story](https://issues.folio.org/browse/UICIRC-405).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@
 * Add Item Limit to Loan Policy. Refs UICIRC-390.
 * Overdue fine policy: Validation field incorrect for overdue recall fine. Refs UIU-1232.
 * Lost item fee policy: Words "late" and "Set cost" missing on entry of new policy. Refs UIU-1315.
+* Remove periods values from default config in `<LoanHistory>`, handle CSS issue when the loan history form is being submitted and add tests for anonymizing intervals. Refs UICIRC-405.
 
 ## [1.11.0](https://github.com/folio-org/ui-circulation/tree/v1.11.0) (2019-09-13)
 [Full Changelog](https://github.com/folio-org/ui-circulation/compare/v1.10.0...v1.11.0)
--- a/src/settings/components/AnonymizingTypeSelect/AnonymizingTypeSelectContainer.js
+++ b/src/settings/components/AnonymizingTypeSelect/AnonymizingTypeSelectContainer.js
@@ -94,7 +94,7 @@ class AnonymizingTypeSelectContainer extends Component {
         <Field
           name={`${path}Selected`}
           component={({ meta }) => {
-            return meta.touched && <span className={css.error}>{meta.error}</span>;
+            return meta.touched && meta.error ? <span className={css.error}>{meta.error}</span> : null;
           }}
         />
       </Fragment>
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@
 * Add Item Limit to Loan Policy. Refs UICIRC-390.
 * Overdue fine policy: Validation field incorrect for overdue recall fine. Refs UIU-1232.
 * Lost item fee policy: Words "late" and "Set cost" missing on entry of new policy. Refs UIU-1315.
+* Remove unnecessary data from `<LoanHistory>` and add tests for anonymizing intervals. Refs UIU-1042.
```
VictorSoroka1: > Remove unnecessary data from `<LoanHistory>`

This is not quite unclear what exactly you have removed, whether unnecessary methods from the component or something else. Please make the message more clear.

test/bigtest/interactors/loan-history/loan-history-form.js
```diff
@@ -21,6 +21,7 @@ import CalloutInteractor from '@folio/stripes-components/lib/Callout/tests/inter
 
   duration = scoped('[data-test-period-duration]', TextFieldInteractor);
   interval = scoped('[data-test-period-interval]', SelectInteractor);
+  selector = scoped('[data-test-period-interval] select');
```
VictorSoroka1: `selector` is not quite clear enough what it is about. Please come up with a more clear name.

VictorSoroka1:
Please also handle CSS issue when the loan history form is being submitted.

src/settings/components/AnonymizingTypeSelect/AnonymizingTypeSelectContainer.js
```diff
@@ -94,7 +94,7 @@ class AnonymizingTypeSelectContainer extends Component {
         <Field
           name={`${path}Selected`}
           component={({ meta }) => {
-            return meta.touched && <span className={css.error}>{meta.error}</span>;
+            return meta.touched && !!meta.error && <span className={css.error}>{meta.error}</span>;
```
VictorSoroka1:
```suggestion
            return meta.touched && meta.error && <span className={css.error}>{meta.error}</span>;
```

SergiyVSergiyenko: Thank you for your comments, Victor. But in this case, since the meta.error by default is undefined, explicit conversion to a boolean is required to avoid rendering an undefined value.

VictorSoroka1: Please see the updated comment. I think it should avoid extra thinking about this converting stuff and thus simplify things from the code readability standpoint.

SergiyVSergiyenko: Yes. Agree. This is better in terms of code readability.

VictorSoroka1: Thank you.

PR #520, opened 2020-02-14
UICIRC-421 - add translations for aria-label values in EditorToolbar

Dmitriy-Litvinenko:
# Purpose
Add translations for aria-label values in EditorToolbar
# Link
https://issues.folio.org/browse/UICIRC-421


VictorSoroka1:
Please also add an entry to the CHANGELOG file.

PR #534, opened 2020-03-25
UICIRC-418: Show in template dropdown only active policies

annamelnyk:
# Description
On New/Edit Patron notice policy page: make Inactive loan and request notice templates **NOT** display in dropdown
# Issue
https://issues.folio.org/browse/UICIRC-418
# Screenshot
**Loan demo (the same with request)**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-circulation
 
+## [2.1.0] (IN PROGRESS)
+
+* New/Edit Patron notice policy: make inactive loan and request notice templates do not display in dropdown. Refs UICIRC-418.
+
 ## 2.0.0 (https://github.com/folio-org/ui-circulation/tree/v2.0.0) (2019-03-13)
 [Full Changelog](https://github.com/folio-org/ui-circulation/compare/v1.12.0...v2.0.0)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Change history for ui-circulation
 
+## [2.1.0] (IN PROGRESS)
+
+* New/Edit Patron notice policy: make inactive loan and request notice templates do not display in dropdown. Refs UICIRC-418.
```
VictorSoroka1:
```suggestion
* New/Edit Patron notice policy: do not display inactive loan and request notice templates in the dropdown. Refs UICIRC-418.
```

src/settings/NoticePolicy/utils/get-templates.js
```diff
@@ -1,8 +1,8 @@
 import { reduce } from 'lodash';
 
 export default (templates = [], categoryName = '') => {
-  return reduce(templates, (items, { id, name, category }) => {
-    if (category === categoryName) {
+  return reduce(templates, (items, { id, name, category, active }) => {
+    if (category === categoryName && active) {
```
VictorSoroka1: I would suggest checking for active ones first. Additionally, it will look a bit neater What do you think?
```suggestion
    if (active && category === categoryName) {
```

annamelnyk: Funny, at first I wrote like you're suggesting and then changed it conversely. But I agree with you.

PR #535, opened 2020-03-26
UICIRC-438 - moving request policy form to react final form

maximdidenkoepam:
# Purpose
Move Request Policy form to React-Final-Form 

# Link
https://issues.folio.org/browse/UICIRC-438


src/settings/RequestPolicy/RequestPolicyForm.js
```diff
@@ -62,19 +60,45 @@ class RequestPolicyForm extends React.Component {
...
+
+    return fetch(`${okapi.url}/request-policy-storage/request-policies?query=(name=="${name}")`,
+      {
+        headers: {
+          'X-Okapi-Tenant': okapi.tenant,
+          'X-Okapi-Token': okapi.token,
+          'Content-Type': 'application/json',
+        }
+      });
+  };
```
zburke, line 74: I hate to nitpick over the names of things, but `checkUniqueName` doesn't "check" anything; rather it returns a list of policies with matching names. If you want it to check, pass in `name` and `initialValues.id` and change the query to retrieve values where the name matches but the id does not. 

Additionally, you can use stripes-connect with a GET mutator to return a promise, same as you have here, without having to fuss with request headers. This is how the `src/settings/Validation/request-policy/unique-name.js` that was removed in this clean-up worked.

maximdidenkoepam: @zburke agree about method name, I've just replaced it. Unfortunately we can't use mutator in this case due to final-form internals. During first render final form will trigger all validation handlers, in case of new policy it seems to be working, but in case of editing mode it will lead to infinite rendering.

zburke: I don't follow about using a mutator. If you set [`accumulate: true` and `fetch: false`](https://github.com/folio-org/ui-users/blob/6aea404b4defdac7fd197c4acfbdad8b812089c0/src/routes/LoanDetailContainer.js#L77-L83) in the manifest, then [`mutator.GET()`](https://github.com/folio-org/ui-users/blob/6aea404b4defdac7fd197c4acfbdad8b812089c0/src/components/Wrappers/withRenew.js#L290-L301) acts just like `fetch()`. ([Documentation](https://github.com/folio-org/stripes-connect/blob/master/doc/api.md#rest-resources))

No worries though; this is a great refactor and I don't want my griping about one query to overshadow that fact.

maximdidenkoepam: @zburke I've implemented validation using approach you proposed in branch UICIRC-439 for patron notice form, and it still behaves I've described above. Could you please take a look and approve or decline approach with fetch for such cases? Thanks.

PR #590, opened 2020-07-01
UICIRC-478 - Lost item fee policy: Additional validation not in original user story

maximdidenkoepam:
# Purpose
To fix validation for Items aged to lost after overdue field.

# Link
https://issues.folio.org/browse/UICIRC-478

### DIFF
--- a/src/settings/Validation/engine/handlers.js
+++ b/src/settings/Validation/engine/handlers.js
@@ -46,7 +46,7 @@ export const hasNoChargeLostItemProcessingFee = (value, model) => {
   const chargeAmountItemPatron = get(model, 'chargeAmountItemPatron');
   const chargeAmountItemSystem = get(model, 'chargeAmountItemSystem');
 
-  return chargeAmountItemPatron !== 'false' || chargeAmountItemSystem !== 'false';
+  return chargeAmountItemPatron || chargeAmountItemSystem;
 };
 
 export const hasLostItemProcessingFeeValue = (value, model) => {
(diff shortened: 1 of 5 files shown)


src/settings/Validation/engine/handlers.js
```diff
@@ -61,3 +60,15 @@ export const hasPositivereplacementAllowed = (value, model) => {
...
+export const hasPatronBilledAfterAgedToLostValue = (value, model) => {
+  const patronBilledAfterAgedToLostValue = get(model, 'patronBilledAfterAgedLost.duration');
+
+  return (isNotEmpty(patronBilledAfterAgedToLostValue) && isNotEmpty(value)) || !isNotEmpty(patronBilledAfterAgedToLostValue);
+};
+
+export const hasChargeAmountItemSystemSelected = (value, model) => {
+  const chargeAmountItemSystem = get(model, 'chargeAmountItemSystem');
+
+  return (chargeAmountItemSystem === 'true' && isNotEmpty(value)) || chargeAmountItemSystem !== 'true';
```
zburke, line 73: We really want to handle these boolean values as strings, i.e. `'true'` not `true`?

maximdidenkoepam: @zburke Field component sends its value in string, even if default was set as boolean. Parse option for Field will store boolean in redux store, but validator will receive string representation. Lost item fee policy was implemented by another team, so I am just following conventions they did.

maximdidenkoepam: @zburke done

PR #612, opened 2020-08-14
UICIRC-464: Fix accessibility errors

annamelnyk:
# Description
Fix of accessibility errors:
- Form elements must have labels
- Buttons must have descernible text
- Certain ARIA roles must be contained to particular parents
- Heading level should only increase by one
- IDs of active elements must be unique
- `id` attribute value must be unique
# Issue
https://issues.folio.org/browse/UICIRC-464


maximdidenkoepam (approved):

test/bigtest/tests/request-policy/request-policy-form-test.js
```diff
@@ -78,9 +78,7 @@ describe('RequestPolicyForm', () => {
         description: 'Request policy 1 desc',
         requestTypes: ['Hold', 'Page']
       });
-    });
 
-    beforeEach(function () {
```
zburke, line 83: Oh interesting! This file has several back-to-back `beforeEach` blocks. I wonder if this is the source of the failing tests I've seen in #602.

annamelnyk: I suppouse it can be the reason, but not sure. Locally tests were passing, but were failing in CI until I removed `beforeEach` blocks

PR #631, opened 2020-09-25
UICIRC-500: Lost item fee displays as Set cost of NaN

annamelnyk:
# Description
In `Lost item fee policy`, when select `Set cost`  and then blank it out (delete the 0.00) -> 
`Lost item fee` displays as `Set cost of NaN`
# Issue
https://issues.folio.org/browse/UICIRC-500
# Screenshot
**Before**
image
**After**
image


maximdidenkoepam (approved):

mkuklis (approved):

zburke (approved):
thank you!

PR #647, opened 2020-10-30
Update feesfines okapi interface

annamelnyk:
Since was made breaking changes in `Modfeesfines` in [PR](https://github.com/folio-org/mod-feesfines/pull/139), we should update version of `feesfines` okapi interface for consistent changes


maximdidenkoepam (approved):

zburke (commented):
If this app no longer supports `v15.0`, please update the major version here as well.

annamelnyk:
Will do @zburke. I also updated version of feesfines in `ui-users`, so I should update the major version of `ui-users` as well right?

zburke (approved):
Technically, yes, if you bump a major version of an interface you must bump the major version of the module that consumes that interface. It's just like bumping the major version of a peer dependency.

PR #660, opened 2020-11-27
UICIRC-528 - Duplicate patron notice template names allowed - case is not considered in validation

maximdidenkoepam:
# Purpose
Make patron notice template name validation case insensitive.

# Link
https://issues.folio.org/browse/UICIRC-528

# Screenshot
Screen Shot 2020-11-27 at 12 34 23

### DIFF
--- a/src/settings/PatronNotices/PatronNoticeForm.js
+++ b/src/settings/PatronNotices/PatronNoticeForm.js
@@ -81,9 +81,11 @@ class PatronNoticeForm extends React.Component {
 
     if (name) {
       try {
-        const response = await this.getTemplatesByName(name);
+        const response = await this.getTemplatesByName(name.toLowerCase());
         const { templates = [] } = await response.json();
-        const matchedTemplate = find(templates, ['name', name]);
+        const matchedTemplate = find(templates, template => {
+          return template.name.toLowerCase() === name.toLowerCase();
+        });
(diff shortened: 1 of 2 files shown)


src/settings/PatronNotices/PatronNoticeForm.js
```diff
@@ -81,9 +81,11 @@ class PatronNoticeForm extends React.Component {
 
     if (name) {
       try {
-        const response = await this.getTemplatesByName(name);
+        const response = await this.getTemplatesByName(name.toLowerCase());
```
annamelnyk, line 84: Here we can to not format string `toLowerCase` since the same is happening below, right?

zburke: The fetch in `this.getTemplatesByName()` is case insensitive (try it yourself for the query in question, `templates?query=(name=="${name}")`**) so `.toLowerCase()` is irrelevant here. It's only the comparison in the `find()` function below that matters, and that looks correct to me. 

** When a field gets a full-text index, an [exact-match search with `==` ignores case](https://github.com/folio-org/raml-module-builder#cql-relations). I can see the logic behind this (e.g. typing in "harry potter") but it surprised me at first because I was thinking of template-names like hash-keys somehow, but really they're not.

zburke (approved):

annamelnyk (approved):

mkuklis (approved):

PR #661, opened 2020-11-27
UICIRC-488: Add space to rule editor to multiple criteria rule

annamelnyk:
# Description
In Circ Rule editor when selecting multiple criteria was a problem with after `Enter` - `rule` with `criteria` were concatenated into regular text. In this PR were added formatted function for adding of `indent` to editor values. Since there were many places where template strings with space used, so I replace them with function too for consistency.
Also add `library` and `campuses` rule multiple selection behavior, because they are - multiple.    
# Task
https://issues.folio.org/browse/UICIRC-488
# Screenshots
**Before:** https://recordit.co/5lroDBsjUm
**After:** https://recordit.co/0xXdNbmq4v

### DIFF
--- a/src/settings/lib/RuleEditor/Rules-hint.js
+++ b/src/settings/lib/RuleEditor/Rules-hint.js
@@ -14,7 +14,7 @@ import {
   RULES_TYPE,
   LOCATION_RULES_TYPES
 } from '../../../constants';
-import { addSpaceToCMRules } from './utils';
+import addIndentToCMRules from './utils';
 
 const locationHeadersMapping = {
   [RULES_TYPE.INSTITUTION]: 'institution',
@@ -64,7 +64,7 @@ function getItemOptions(selector, typeKey) {
   const displayText = getDisplayText(selector, typeKey);
 
   return {
-    text: addSpaceToCMRules(text, 'after'),
+    text: addIndentToCMRules(text, 'after'),
     displayText,
     className: 'rule-hint-minor',
     id: selector.id,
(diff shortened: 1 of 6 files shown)


maximdidenkoepam (approved):

src/settings/lib/RuleEditor/Rules-hint.js
```diff
@@ -63,7 +64,7 @@ function getItemOptions(selector, typeKey) {
   const displayText = getDisplayText(selector, typeKey);
 
   return {
-    text: `${text} `,
+    text: addIndentToCMRules(text, 'after'),
```
ViktorSoroka07, line 67: Could you please clarify what CM means?

annamelnyk: Sure. CM stands for [CodeMirror](https://codemirror.net/) - the name of text editor which is used as Circulation Rules Editor.

mkuklis (approved):

src/settings/lib/RuleEditor/Rules-hint.js
ViktorSoroka07: I think it would be better to omit this in the name for better readability or write as you mentioned it `CodeMirror`.

annamelnyk: Agree. Changed func name to `addIndentToEditorRules`.

ViktorSoroka07 (approved):

PR #688, opened 2021-01-28
UICIRC-513: Move moment from a regular to a peer and dev dependencies

Dmitriy-Litvinenko:
# Purpose
Move moment from a regular to a peer and dev dependencies.

# Link
https://issues.folio.org/browse/UICIRC-513


annamelnyk (approved):

zburke (approved):
Thank you for cleaning this up. I know it seems small, but sanding off these rough edges really does make everything better.

PR #715, opened 2021-04-14
UICIRC-563 also support circulation 10.0

zburke:
The only breaking change in `circulation` `10.0` was the removal of the
`override-check-out-by-barcode` which is not used here, hence continuing
to support `9.0`.

Refs [UICIRC-563](https://issues.folio.org/browse/UICIRC-563)


Dmitriy-Litvinenko (approved):

package.json
```diff
@@ -29,7 +29,7 @@
       "@folio/stripes-template-editor"
     ],
     "okapiInterfaces": {
-      "circulation": "9.0",
+      "circulation": "9.0 10.0",
```
Dmitriy-Litvinenko, line 32: Hello @zburke thank you
I may be wrong, but override-check-out-by-barcode was only used in ui-checkout and not used in other folio ui modules. override-check-out-by-barcode was removed from ui-checkout in Q12021 after "migration" to other endpoint.
Could you please clarify how i understood that we can migrate from version 9 to version 10 without support previous version ("circulation": "10.0",)?

PR #723, opened 2021-04-23
UICIRC-568: Fix limit in policy settings

mkuklis:
The `limit` was already set to correct value `1000` but it was overwritten by `perRequest`.

This PR removes `perRequest` and replaces `limit` with `maxUnpagedResourceCount` introduced by @MikeTaylor in: https://github.com/folio-org/ui-inventory/pull/1359

`maxUnpagedResourceCount` is set to `1000` by default.

https://issues.folio.org/browse/UICIRC-568

### DIFF
--- a/src/settings/FinePolicy/FinePolicySettings.js
+++ b/src/settings/FinePolicy/FinePolicySettings.js
@@ -13,17 +13,17 @@ import FinePolicy from '../Models/FinePolicy';
 import FinePolicyDetail from './FinePolicyDetail';
 import FinePolicyForm from './FinePolicyForm';
 import normalize from './utils/normalize';
+import { MAX_UNPAGED_RESOURCE_COUNT } from '../../constants';
 
 class FinePolicySettings extends React.Component {
   static manifest = Object.freeze({
     finePolicies: {
       type: 'okapi',
       records: 'overdueFinePolicies',
-      perRequest: 100,
       path: 'overdue-fines-policies',
       params: {
         query: 'cql.allRecords=1',
-        limit: '1000',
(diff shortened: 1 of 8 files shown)


src/settings/FinePolicy/FinePolicySettings.js
```diff
@@ -13,17 +13,17 @@ import FinePolicy from '../Models/FinePolicy';
...
   static manifest = Object.freeze({
     finePolicies: {
       type: 'okapi',
       records: 'overdueFinePolicies',
-      perRequest: 100,
       path: 'overdue-fines-policies',
       params: {
         query: 'cql.allRecords=1',
-        limit: '1000',
+        limit: (q, p, r, l, props) => props?.stripes?.config?.maxUnpagedResourceCount || MAX_UNPAGED_RESOURCE_COUNT,
```
MikeTaylor, line 26: This looks good to me.

(If I had my life all over again, these callback functions would _only_ pass props, and if you wanted query, path, etc. you'd extract them from the props. But that's for a new version of stripes-connect in the other leg of the trousers of time.)

src/settings/PatronNotices/PatronNotices.js
```diff
@@ -53,7 +56,7 @@ class PatronNotices extends React.Component {
       path: 'patron-notice-policy-storage/patron-notice-policies',
       params: {
         query: 'cql.allRecords=1',
-        limit: '1000',
+        limit: (q, p, r, l, props) => props?.stripes?.config?.maxUnpagedResourceCount || MAX_UNPAGED_RESOURCE_COUNT,
```
MikeTaylor, line 59: Wow, you're using this _a lot_!

MikeTaylor (approved):
Nice.

zburke (approved):

PR #782, opened 2021-09-15
UICIRC-655: UI tests replacement with RTL/Jest for folder Period

pavelspichonak:
## Purpose
Add RTL/Jest unit tests for folder Period

## Refs
https://issues.folio.org/browse/UICIRC-655

## Screenshots
Screen Shot 2021-09-17 at 11 15 11 AM

### DIFF
--- a/src/settings/components/Period/Period.js
+++ b/src/settings/components/Period/Period.js
@@ -18,9 +18,17 @@ import {
   TextField,
 } from '@folio/stripes/components';
 
+export const transformInputValue = (value) => {
+  if (isEmpty(value)) {
+    return '';
+  }
+
+  return Number(value);
+};
+
 class Period extends PureComponent {
   static propTypes = {
-    intl: PropTypes.object,
+    intl: PropTypes.object.isRequired,
     inputValuePath: PropTypes.string.isRequired,
     selectValuePath: PropTypes.string.isRequired,
     intervalPeriods: PropTypes.arrayOf(PropTypes.node).isRequired,
@@ -54,14 +62,6 @@ class Period extends PureComponent {
     changeFormValue(inputValuePath, '');
   };
 
-  transformInputValue = (value) => {
-    if (isEmpty(value)) {
-      return '';
-    }
-
-    return Number(value);
-  };
-
   render() {
     const {
       fieldLabel,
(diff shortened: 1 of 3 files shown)


src/settings/components/Period/Period.test.js
```diff
@@ -0,0 +1,195 @@
...
+        {...mockedProps}
+        intervalPeriodsSuffix={labelIds.intervalPeriodsSuffixId}
+      />
+    );
+
+    const periodIntervalColumn = screen.getByTestId('periodIntervalColumn');
+    expect(periodIntervalColumn).toHaveAttribute('xs', '3');
+  });
+
+  it('should render component when "intervalPeriodsSuffix" props is passed', () => {
```
Stanislau-Snapok, line 170: we can swap this test and the previous one. in this case we first check correct render `intervalPeriodsSuffix` section when corresponding property is passed, and then we can render component with 'intervalPeriodsSuffix' prop, and with 'selectSize' which is differ from default. so we can avoid step with rerender and just check that both sections have correct 'xs' attribute.

Stanislau-Snapok (commented):
Add please info in CHANGELOG.md
And may be we can do description of tests more accurate? For example we can wright something like "should render 'periodIntervalColumn' section ..." instead of  "should render component ...". It will gave us more accurate information where we must look for mistake, if something whent wrong.

src/settings/components/Period/Period.js
```diff
@@ -82,20 +82,18 @@ class Period extends PureComponent {
...
+                <Label required={required}>
+                  <FormattedMessage id={fieldLabel} />
+                </Label>
               </Col>
             </Row>
           </div>
         )}
         <Row>
-          <Col xs={inputSize}>
+          <Col data-testid="finePolicyDurationColumn" xs={inputSize}>
```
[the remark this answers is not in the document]
pavelspichonak: done

src/settings/components/Period/Period.test.js
```diff
@@ -0,0 +1,195 @@
...
+  render,
+  screen,
+  within,
+  fireEvent,
+} from '@testing-library/react';
+
+import '../../../../test/jest/__mock__';
+
+import { Field } from 'react-final-form';
+import { TextField, Select } from '@folio/stripes/components';
```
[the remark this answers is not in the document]
pavelspichonak: done

src/settings/components/Period/Period.test.js
pavelspichonak: done. thanks

src/settings/components/Period/Period.js
```diff
@@ -82,20 +82,24 @@ class Period extends PureComponent {
...
       <>
         {fieldLabel && (
-          <div data-test-period-label>
+          <div
+            data-test-period-label
+            data-testid="periodLabelSection"
+          >
             <Row>
               <Col xs={12}>
-                { fieldLabel && (
```
pavelspichonak, line 88: unnecessary condition. we have the same above

Stanislau-Snapok (approved):

Dmitriy-Litvinenko (approved):

PR #789, opened 2021-09-20
UICIRC-430 improve location display in circ rules

zburke:
The CSS rules governing the line-height of a single location-item are
completely impenetrable to me, but removing the `height` attribute on
the container element somehow, mysteriously, fixes the display. It
appears that what's happening is a bunch of block-level elements with a
too-short line-height are stacked, such that the line below clips that
above, but I couldn't figure out how to change the line-height for a
single element anywhere.

Refs [UICIRC-430](https://issues.folio.org/browse/UICIRC-430)

### DIFF
--- a/src/settings/lib/RuleEditor/CodeMirrorCustom.css
+++ b/src/settings/lib/RuleEditor/CodeMirrorCustom.css
@@ -179,7 +179,7 @@
 }
 
 .CodeMirror-hints-list ul li {
-  height: 12px;
+  height: auto;
 }
 
 .CodeMirror-hints-list.CodeMirror-hints-multiple-selection ul {
--- a/src/settings/lib/RuleEditor/CodeMirrorCustom.css
+++ b/src/settings/lib/RuleEditor/CodeMirrorCustom.css
@@ -90,9 +90,9 @@
   width: 100%;
 }
 
+/* wrapper for the _entire_ list, not a single list-item */
 .CodeMirror-checkbox-container {
   width: 100%;
-  height: 12px;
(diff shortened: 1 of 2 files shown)


Dmitriy-Litvinenko:
@zburke thank you 
I have one question
image
Can we add style for fix selected value? Just for example:
```
.CodeMirror-hints-list ul li {
  height: auto!important
}
```
image

src/settings/lib/RuleEditor/CodeMirrorCustom.css
```diff
@@ -90,9 +90,9 @@
   width: 100%;
 }
 
+/* wrapper for the _entire_ list, not a single list-item */
 .CodeMirror-checkbox-container {
   width: 100%;
-  height: 12px;
```
Dmitriy-Litvinenko, line 95: @zburke thank you 
I have one question
image
Can we add style for fix focused/selected value? Just for example:
```
.CodeMirror-hints-list ul li {
  height: auto!important
}
```
image

zburke: Nice find! I don't think we even need the `!important`, which I try to avoid as a rule. We can just change [the existing style on line 182](https://github.com/folio-org/ui-circulation/blob/fee7f30f799e0190d08e8f70d2a05f7630ed054f/src/settings/lib/RuleEditor/CodeMirrorCustom.css#L181-L183) from `12px` to `auto`.

Dmitriy-Litvinenko (approved):

PR #790, opened 2021-09-21
UICIRC-692 provide view-only perms for overdue, lost-item policies

zburke:
1. Provide read-only permissions for overdue fine policies and lost item
   fee policies.
2. Redefine the existing CRUD permissions in terms of the read-only
   permission set.
3. Allow users with read-only permission to view overdue and lost-item
   policies. Previously, full CRUD access was required.

Refs UICIRC-692


src/index.js
```diff
@@ -80,13 +80,13 @@ class Circulation extends Component {
             route: 'fine-policies',
             label: <FormattedMessage id="ui-circulation.settings.finePolicy.paneTitle" />,
             component: FinePolicySettings,
-            perm: 'ui-circulation.settings.overdue-fines-policies',
+            perm: 'ui-circulation.settings.read-overdue-fines-policies',
```
Dmitriy-Litvinenko, line 83: Could you please clarify this is correct that we create ui-circulation.settings.**view**-overdue-fines-policies but we check ui-circulation.settings.**read**-overdue-fines-policies
https://github.com/folio-org/ui-circulation/pull/790/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R221
The same below

zburke: Nope, that's my error! Good catch, thank you. Fixed. (I thought of these as "read" policies and drafted them that way, but the ticket refers to them as "view" policies, so I changed the language but missed a spot.)

annamelnyk (approved):

package.json
```diff
@@ -217,15 +217,35 @@
...
+        "visible": true
+      },
       {
         "permissionName": "ui-circulation.settings.overdue-fines-policies",
         "displayName": "Settings (Circ): Can create, edit and remove overdue fine policies",
         "subPermissions": [
+          "ui-circulation.settings.view-overdue-fines-policies",
           "overdue-fines-policies.item.post",
           "overdue-fines-policies.item.put",
-          "overdue-fines-policies.item.delete",
```
Dmitriy-Litvinenko, line 226: Could you please confirm that we need delete "overdue-fines-policies.item.delete" permissions here?

package.json
```diff
@@ -236,14 +256,10 @@
         "permissionName": "ui-circulation.settings.lost-item-fees-policies",
         "displayName": "Settings (Circ): Can create, edit and remove lost item fee policies",
         "subPermissions": [
+          "ui-circulation.settings.view-lost-item-fees-policies",
           "lost-item-fees-policies.item.post",
           "lost-item-fees-policies.item.put",
-          "lost-item-fees-policies.item.delete",
```
Dmitriy-Litvinenko, line 241: Could you please confirm that we need delete "lost-item-fees-policies.item.delete" permissions here?

zburke: I did not make any substantive changes to the "Can create, edit and remove..." policy, only restructured it so the "view" values are now pulled in indirectly via the "view-only" policy. It appears as changed lines only because `delete` is now the last value and therefore has no trailing comma.

Note [line 262](https://github.com/folio-org/ui-circulation/blob/0c5138227812f0747cef4c0ef1932b79b1a8f4da/package.json#L262)

Dmitriy-Litvinenko (approved):

PR #795, opened 2021-09-27
UI tests replacement with RTL/Jest for component SchedulesList (EditS…

pavelspichonak:
## Purpose
UI tests replacement with RTL/Jest for component SchedulesList

## Refs
https://issues.folio.org/browse/UICIRC-603

## Screenshots
Screen Shot 2021-09-27 at 11 13 07 AM

### DIFF
--- a/src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
+++ b/src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
@@ -1,0 +1,184 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  within,
+  fireEvent,
+} from '@testing-library/react';
+import {
+  isArray,
+  uniqueId,
+} from 'lodash';
+
+import '../../../../../../../test/jest/__mock__';
+
+import SchedulesList from './SchedulesList';
+import ScheduleCard from '../ScheduleCard';
+
+jest.mock('../ScheduleCard', () => jest.fn(({
+  onRemoveSchedule,
+  scheduleIndex,
+}) => (
+  // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+  <div
+    data-testid="scheduleCard"
+    onClick={() => onRemoveSchedule(scheduleIndex)}
+  />
+)));
+jest.mock('lodash', () => {
+  const originalModule = jest.requireActual('lodash');
+
+  return {
+    __esModule: true,
+    ...originalModule,
+    uniqueId: jest.fn((prefix) => `${prefix}_uniqueId`),
+  };
+});
+
+describe('SchedulesList', () => {
+  const testTimeZone = 'UTC';
+  const testFieldsNames = [
+    'testSchedule1',
+    'testSchedule2',
+  ];
+  const testFieldsValue = [
+    {
+      key: 'testScheduleKey1',
+    },
+    {
+      key: 'testScheduleKey2',
+    },
+  ];
+  const testFields = {
+    push: jest.fn(),
+    remove: jest.fn(),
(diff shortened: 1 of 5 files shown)


src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
```diff
@@ -0,0 +1,186 @@
...
+      key: 'testScheduleKey1',
+    },
+    {
+      key: 'testScheduleKey2',
+    },
+  ];
+  const testFields = {
+    push: jest.fn(),
+    remove: jest.fn(),
+    map: (callback) => testFieldsNames.map((name, index) => callback(name, index)),
```
Stanislau-Snapok, line 55: map: (callback) => testFieldsNames.map(callback) - should work, i think

src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
```diff
@@ -0,0 +1,186 @@
...
+      renderComponent();
+    });
+
+    it('should render "add schedule" button', () => {
+      expect(within(screen.getByTestId('addScheduleButton')).getByText(labelIds.new))
+        .toBeVisible();
+    });
+
+    it('should add schedule', () => {
+      const addScheduleButton = screen.getByTestId('addScheduleButton');
```
Stanislau-Snapok, line 91: we can get it by role, so we don't need to add test id for it.

src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
```diff
@@ -0,0 +1,186 @@
...
+      expect(testFields.push).toHaveBeenCalledWith({ key: uniqueId.mock.results[0].value });
+    });
+  });
+
+  describe(('schedule cards'), () => {
+    beforeEach(() => {
+      renderComponent();
+    });
+
+    it('should render schedule cards', () => {
```
Stanislau-Snapok, line 105: i think it will be better to extract this logic into separate function, and check each item separatly. additionaly i would extract "should remove schedule" test into test function too.

annamelnyk (approved):

src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
pavelspichonak: good point. done

src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
pavelspichonak: we can but in my opinion it is better to be more specific. for example in case when we use `data-testid` and some additional button is added we will have no problems in our tests. But if we use `getByRole` our test fail because there will be two buttons on the page.

src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
pavelspichonak: Fixed but let us discuss

src/settings/FixedDueDateSchedule/components/EditSections/components/SchedulesList/SchedulesList.test.js
Stanislau-Snapok: ok, accepted.

Stanislau-Snapok (approved):

Dmitriy-Litvinenko (approved):

PR #796, opened 2021-09-27
UICIRC-645: Add tests for 'PatronNotices/TokensList/TokensList.js'

annamelnyk:
## Purpose
Add RTL/Jest testing for `TokensList` component in `PatronNotices/TokensList/TokensList.js`

## Refs
https://issues.folio.org/browse/UICIRC-645

## Screenshots
image

## Pre-Merge Checklist
Before merging this PR, please go through the following list and take appropriate actions.
- [x] I've added appropriate record to the CHANGELOG.md
- Does this PR meet or exceed the expected quality standards?
  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] If any API-related changes - okapi interfaces and permissions are reviewed/changed correspondingly
  - [x] There are no breaking changes in this PR.

If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.


mkuklis (approved):

src/settings/PatronNotices/TokensList/TokensList.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+      expect(sectionTestIdValue).toBeVisible();
+      expect(sectionTestIdValue).toHaveAttribute('section', sectionTestId);
+      expect(within(screen
+        .getByTestId(`headerFor${sectionTestId}`))
+        .getByText(`ui-circulation.settings.patronNotices.${sectionTestId}TokenHeader`))
+        .toBeVisible();
+    });
+
+    tokensProps[sectionTestId].map((token) => {
```
Stanislau-Snapok, line 80: for what we are doing that? i tried start tests, and never see `should render checkbox...` test. how `tokensProps` even connected with `TokensList` component? didn't find it in component.

src/settings/PatronNotices/TokensList/TokensList.test.js
```diff
@@ -0,0 +1,130 @@
...
+  render,
+  within,
+} from '@testing-library/react';
+
+import '../../../../test/jest/__mock__';
+
+import tokensProps from '../tokens';
+import TokensList from './TokensList';
+
+jest.mock('@folio/stripes-template-editor', () => ({
```
Stanislau-Snapok, line 14: i think we should not mock `TokensSection` like this. jest.fn(() => null) will be enough for us in my opinion. then we can check that each `TokensSection` was executed with the correct props. the only nuance for label id check we should inject intl and replace `FormattedMessage` component with `formatMessage` from intl if the component allows us. if not -  just mock something like jest.fn(({header}) => <div>{header}</div>), and check each `TokensSection` execution with expect.objectContaining(props).

annamelnyk: The `TokensList` component represents list of tokens(checkboxes) broke into sections, and I suppose would be appropriate to test it deeply.

src/settings/PatronNotices/TokensList/TokensList.test.js
annamelnyk: `TokensList` component receives list of tokens as prop, so here also `tokensProps` are tested.
And in callback was missed destructuring of `token`, because of this you haven't met `should render checkbox`.
Fixed that already. Thanks for review.

src/settings/PatronNotices/TokensList/TokensList.test.js
Stanislau-Snapok: Ok, now i got it. But in my mind render of checkboxes the business of `TokensSection`, and now it's mocked. So, do we realy need to test our mock? I think better test `TokensSection` separately. And in our case, since it is a stripes component, it means never.

src/settings/PatronNotices/TokensList/TokensList.test.js
Stanislau-Snapok: If in the end this approach will be approved, i think it would be better to use `forEach`, because in this case we don't need to return a new modified array.

src/settings/PatronNotices/TokensList/TokensList.test.js
zburke: These are thoughtful questions, @Stanislau-Snapok! I see how this test is written to cover `<TokensList>`, but also agree with you that much of it appears to test the mock of `<TokensSection>`. Your are correct, but in the end it does serve to cover `<TokensList>` as well. 🤷  

I think this is less "test is sub-optimal" than "patterns in our codebase may not make testing easy." I really appreciate your comments here. They encourage us all to think about how to write good tests, and how to write easily testable code.

zburke (approved):

PR #797, opened 2021-09-28
UICIRC-615: UI tests replacement with RTL/Jest for component RequestM…

pavelspichonak:
## Purpose
UI tests replacement with RTL/Jest for component RequestManagementSection

## Refs
https://issues.folio.org/browse/UICIRC-615

## Screenshots
Screen Shot 2021-09-28 at 1 42 24 PM

### DIFF
--- a/src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.test.js
+++ b/src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.test.js
@@ -1,0 +1,397 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  within,
+} from '@testing-library/react';
+
+import '../../../../../../test/jest/__mock__';
+
+import { Field } from 'react-final-form';
+import {
+  Checkbox,
+  Accordion,
+} from '@folio/stripes/components';
+import {
+  intervalPeriods,
+  loanProfileMap,
+} from '../../../../../constants';
+import Period from '../../../../components/Period';
+import RequestManagementSection from './RequestManagementSection';
+
+jest.mock('../../../../utils/options-generator', () => jest.fn((
+  formatMessage,
+  periods,
+  label,
+) => ({
+  periods,
+  label,
+})));
+
+jest.mock('../../../../components/Period', () => jest.fn(() => null));
+
+describe('RequestManagementSection', () => {
+  const labelIds = {
+    requestManagement: 'ui-circulation.settings.requestManagement.requestManagement',
+    recalls: 'ui-circulation.settings.requestManagement.recalls',
+    holds: 'ui-circulation.settings.requestManagement.holds',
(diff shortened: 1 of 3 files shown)


src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.test.js
```diff
@@ -0,0 +1,456 @@
...
+            loanable: true,
+          }}
+        />
+      );
+
+      expect(Period).toHaveBeenNthCalledWith(
+        periodCallOrderByPlace.minimumGuaranteedLoan,
+        {
+          fieldLabel: labelIds.minimumGuaranteedLoanPeriod,
+          inputValuePath: 'requestManagement.recalls.minimumGuaranteedLoanPeriod.duration',
```
Stanislau-Snapok, line 225: i think this one and next three props can be moced into separate constant value. you use them at least four times.

src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.test.js
```diff
@@ -0,0 +1,456 @@
...
+          type: 'checkbox',
+        }, {}
+      );
+    });
+  });
+
+  describe('"renew items with request" field', () => {
+    it('should render field', () => {
+      render(
+        <RequestManagementSection
```
Stanislau-Snapok, line 436: if i dont mistake, you render `RequestManagementSection` with this set of props at least five times. so can we put the cases with same render in one additional `describe` and use `beforeEach` construction? what do you think about it?

Stanislau-Snapok (approved):

src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.test.js
Dmitriy-Litvinenko: Could you please move 
```
intervalPeriods: intervarPeriodsExpectedValue,
changeFormValue: changeMock,
```
to variable and reuse it across all file?

pavelspichonak: done

src/settings/LoanPolicy/components/EditSections/RequestManagementSection/RequestManagementSection.test.js
pavelspichonak: I already have such describe `when "policy.loanable" is true` but to be honest I would not like to move these tests there. The idea is to test all components separately. We can discuss it.

pavelspichonak: done

PR #799, opened 2021-09-28
UICIRC-650: Add RTL/Jest tests for `StaffSlipForm` component

Stanislau-Snapok:
## Purpose
Add RTL/Jest testing for `StaffSlipForm` component in `settings/StaffSlips`

## Refs
https://issues.folio.org/browse/UICIRC-650

## Screenshots
image

### DIFF
--- a/src/settings/StaffSlips/StaffSlipForm.test.js
+++ b/src/settings/StaffSlips/StaffSlipForm.test.js
@@ -1,0 +1,202 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  fireEvent,
+} from '@testing-library/react';
+
+import '../../../test/jest/__mock__';
+
+import { Field } from 'react-final-form';
+
+import { TemplateEditor } from '@folio/stripes-template-editor';
+import {
+  KeyValue,
+  Pane,
+  Paneset,
+  TextArea,
+} from '@folio/stripes/components';
+
+import StaffSlipForm from './StaffSlipForm';
+import TokensList from './TokensList';
+import tokens from './tokens';
+import {
+  CancelButton,
+  FooterPane,
+} from '../components';
+
+jest.mock('@folio/stripes/final-form', () => jest.fn(() => (component) => component));
+jest.mock('../components', () => ({
+  CancelButton: jest.fn(() => null),
+  FooterPane: jest.fn(() => null),
+}));
+Field.mockImplementation(jest.fn(() => null));
+Pane.mockImplementation(jest.fn(({ children, firstMenu, footer }) => (
+  <div>
+    {firstMenu}
+    {children}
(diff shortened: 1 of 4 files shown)


src/settings/StaffSlips/StaffSlipForm.test.js
```diff
@@ -0,0 +1,201 @@
...
+    description: 'ui-circulation.settings.staffSlips.description',
+    display: 'ui-circulation.settings.staffSlips.display',
+    preview: 'ui-circulation.settings.staffSlips.form.previewLabel',
+    new: 'ui-circulation.settings.staffSlips.new',
+  };
+  const mockedHandleSubmit = jest.fn();
+  const mockedOnCancel = jest.fn();
+
+  afterEach(() => {
+    Field.mockClear();
```
pavelspichonak, line 58: nit: maybe to clear mock for `Pane` as well? Not sure if it is necessary here but just in case.

pavelspichonak (approved):
LGTM

src/settings/StaffSlips/StaffSlipForm.test.js
```diff
@@ -0,0 +1,201 @@
...
+  CancelButton,
+  FooterPane,
+} from '../components';
+
+jest.mock('@folio/stripes/final-form', () => jest.fn(() => (component) => component));
+jest.mock('../components', () => ({
+  CancelButton: jest.fn(() => null),
+  FooterPane: jest.fn(() => null),
+}));
+Field.mockImplementation(jest.fn(() => null));
```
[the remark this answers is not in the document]
Stanislau-Snapok: Yeah, in this case we need it, because if we don't do that, we will have tests fail with error `Cannot call a class as a function`. I tried to modify general mock, but in this case some other tests failed.

src/settings/StaffSlips/StaffSlipForm.test.js
Stanislau-Snapok: In this component we have only one `Pane`, and in all cases i check it with `toHaveBeenLastCalledWith` method, so it's not necessary in my opinion, but if you sleep better, I'll add it before merge =)

src/settings/StaffSlips/StaffSlipForm.test.js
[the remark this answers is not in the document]
pavelspichonak: I see, ok

src/settings/StaffSlips/StaffSlipForm.test.js
pavelspichonak: it is okay, it is up to you

Dmitriy-Litvinenko (approved):

PR #805, opened 2021-10-06
UICIRC-699: Add Title Level Request permission

Stanislau-Snapok:
## Purpose
Add Title Level Request permission

## Refs
https://issues.folio.org/browse/UICIRC-699


Dmitriy-Litvinenko (approved):

translations/ui-circulation/en.json
```diff
@@ -469,5 +469,6 @@
   "permission.settings.overdue-fines-policies": "Settings (Circ): Can create, edit and remove overdue fine policies",
   "permission.settings.lost-item-fees-policies": "Settings (Circ): Can create, edit and remove lost item fee policies",
   "permission.settings.view-overdue-fines-policies": "Settings (Circ): Can view overdue fine policies",
-  "permission.settings.view-lost-item-fees-policies": "Settings (Circ): Can view lost item fee policies"
+  "permission.settings.view-lost-item-fees-policies": "Settings (Circ): Can view lost item fee policies",
+  "permission.settings.titleLevelRequests": "Settings (Circ): Can modify title level requests settings"
```
pavelspichonak, line 473: it looks a little bit weird that we have camelcase and snakecase in translations. Looks like different modules have different agreements.

Dmitriy-Litvinenko: @pavelspichonak  How i understood we don't have rules that describe name convention for current case. On my opinion using camel case here will be better and this will be consistency with naming that we use in translation files.
CC: @Stanislau-Snapok

PR #806, opened 2021-10-06
UICIRC-623: UI tests replacement with RTL/Jest for component LostItem…

pavelspichonak:
## Purpose
UI tests replacement with RTL/Jest for component LostItemFeeSection

## Refs
https://issues.folio.org/browse/UICIRC-623

## Screenshots
Screen Shot 2021-10-06 at 3 13 07 PM

### DIFF
--- a/src/settings/LostItemFeePolicy/components/EditSections/LostItemFeeSection/LostItemFeeSection.test.js
+++ b/src/settings/LostItemFeePolicy/components/EditSections/LostItemFeeSection/LostItemFeeSection.test.js
@@ -1,0 +1,465 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  within,
+} from '@testing-library/react';
+
+import '../../../../../../test/jest/__mock__';
+
+import { Field } from 'react-final-form';
+import {
+  Select,
+  Label,
+  RadioButton,
+  TextField,
+  Accordion,
+} from '@folio/stripes/components';
+import {
+  intervalPeriodsLower,
+} from '../../../../../constants';
+import RadioGroup from '../RadioGroup/RadioGroup';
+import Period from '../../../../components/Period';
+import LostItemFeeSection, {
+  formatNumber,
+} from './LostItemFeeSection';
+
+jest.mock('../../../../utils/options-generator', () => jest.fn((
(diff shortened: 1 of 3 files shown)


src/settings/LostItemFeePolicy/components/EditSections/LostItemFeeSection/LostItemFeeSection.test.js
```diff
@@ -0,0 +1,474 @@
...
+      describe(`when "lostItemFeeSectionOpen" prop is ${lostItemFeeSectionOpen}`, () => {
+        it('should render accordion', () => {
+          render(
+            <LostItemFeeSection
+              {...defaultTestProps}
+              lostItemFeeSectionOpen={lostItemFeeSectionOpen}
+            />
+          );
+
+          expect(Accordion).toHaveBeenNthCalledWith(
```
Stanislau-Snapok, line 448: nit: use `toHaveBeenLastCalledWith` instead of `toHaveBeenNthCalledWith` and remove order map for `Accordion`.

src/settings/LostItemFeePolicy/components/EditSections/LostItemFeeSection/LostItemFeeSection.test.js
```diff
@@ -0,0 +1,474 @@
...
+      );
+    });
+
+    it('should render radio group', () => {
+      expect(RadioGroup).toHaveBeenCalled();
+    });
+
+    it('should render "chargeAmountItemPatron" text', () => {
+      expect(
+        within(screen.getByTestId('chargeAmountItemPatronColumn'))
```
Stanislau-Snapok, line 393: we can use `const getById = (id) => within(screen.getByTestId(id))`, and then just getById(id).getByText(text). and as you mentioned earlie, when we will extract `getById` in more global level, we can just remove function definition and add import without any renaming. and i think we should do extracting asap.
CC: @Dmitriy-Litvinenko

pavelspichonak: agree. done

PR #809, opened 2021-10-08
UICIRC-641: UI tests replacement with RTL/Jest for component NoticesList (EditSections)

Dmitriy-Litvinenko:
## Purpose
Add RTL/Jest tests for component NoticesList (EditSections)

## Refs
https://issues.folio.org/browse/UICIRC-641

### DIFF
--- a/src/settings/NoticePolicy/components/EditSections/components/NoticesList/NoticesList.test.js
+++ b/src/settings/NoticePolicy/components/EditSections/components/NoticesList/NoticesList.test.js
@@ -1,0 +1,154 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  within,
+  fireEvent,
+} from '@testing-library/react';
+
+import '../../../../../../../test/jest/__mock__';
+
+import NoticesList from './NoticesList';
+import NoticeCard from '../NoticeCard';
+
+jest.mock('../NoticeCard', () => jest.fn(({
+  onRemoveNotice,
+  noticeIndex,
+}) => (
+  // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+  <div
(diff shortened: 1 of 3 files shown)


src/settings/NoticePolicy/components/EditSections/components/NoticesList/NoticesList.test.js
```diff
@@ -0,0 +1,152 @@
+import React from 'react';
+import {
+  render,
+  screen,
+  within,
+} from '@testing-library/react';
+import fireEvent from '@testing-library/user-event';
```
Stanislau-Snapok, line 7: I am sure that we can import `fireEvent` directly from `@testing-library/react`

src/settings/NoticePolicy/components/EditSections/components/NoticesList/NoticesList.test.js
```diff
@@ -0,0 +1,152 @@
...
+    testFields.remove.mockClear();
+  });
+
+  describe(('notice cards'), () => {
+    beforeEach(() => {
+      renderComponent();
+    });
+
+    const generateTests = (testFieldName, noticeIndex) => {
+      describe((`notice card #${noticeIndex}`), () => {
```
Stanislau-Snapok, line 104: Do we really need `describe` with "...#0"? Or should it start from "1" for example? If it is, we can extract it in constant, and also use it in `toHaveBeenNthCalledWith` as call order number.

Stanislau-Snapok (approved):

src/settings/NoticePolicy/components/EditSections/components/NoticesList/NoticesList.test.js
Dmitriy-Litvinenko: Fixed

PR #810, opened 2021-10-12
UICIRC-643: UI tests replacement with RTL/Jest for folder PatronNotic…

pavelspichonak:
## Purpose
UI tests replacement with RTL/Jest for folder PatronNoticeDetail

## Refs
https://issues.folio.org/browse/UICIRC-643

## Screenshots
Screen Shot 2021-10-12 at 1 01 14 PM

### DIFF
--- a/src/settings/PatronNotices/PatronNoticeDetail.js
+++ b/src/settings/PatronNotices/PatronNoticeDetail.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import HtmlToReact, { Parser } from 'html-to-react';
-import { FormattedMessage } from 'react-intl';
+import { injectIntl } from 'react-intl';
 import { get } from 'lodash';
 
 import {
@@ -19,6 +19,7 @@ import tokens from './tokens';
 class PatronNoticeDetail extends React.Component {
   static propTypes = {
     initialValues: PropTypes.object.isRequired,
+    intl: PropTypes.object.isRequired,
   };
 
   constructor(props) {
@@ -59,7 +60,10 @@ class PatronNoticeDetail extends React.Component {
   }
 
   render() {
-    const { initialValues: notice } = this.props;
+    const {
+      initialValues: notice,
+      intl: { formatMessage },
+    } = this.props;
     const {
       accordions,
       openPreview,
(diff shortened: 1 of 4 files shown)


src/settings/PatronNotices/PatronNoticeDetail.js
```diff
@@ -114,18 +118,21 @@ class PatronNoticeDetail extends React.Component {
...
                       value={notice.localizedTemplates.en.header}
                     />
                   </Col>
-                  <Col xs={4}>
+                  <Col
+                    data-testid="previewButtonColumn"
+                    xs={4}
+                  >
                     <Button onClick={this.openPreviewDialog}>
                       <FormattedMessage id="ui-circulation.settings.patronNotices.preview" />
```
Stanislau-Snapok, line 137: Can we here replace `FormattedMessage` by `formatMessage` too, and remove `FormattedMessage` from imports as unnecessary?

src/settings/PatronNotices/PatronNoticeDetail.test.js
```diff
@@ -0,0 +1,312 @@
...
+  default: {
+    ProcessNodeDefinitions: jest.fn(() => ({
+      processDefaultNode: 'defaultNode',
+    })),
+  },
+  Parser: () => ({
+    parseWithInstructions: jest.fn(() => mockParseWithInstructionsReturnValue),
+  }),
+}));
+Button.mockImplementation(({ children, onClick }) => (
```
Stanislau-Snapok, line 45: I believe that global `Button` mock will be enough here. It also render `children` and have `onClick` property.

src/settings/PatronNotices/PatronNoticeDetail.test.js
```diff
@@ -0,0 +1,312 @@
...
+  });
+
+  [
+    false,
+    true,
+  ].forEach((active) => {
+    describe(`when 'active' is ${active}`, () => {
+      beforeEach(() => {
+        const props = {
+          ...defaultTestProps,
```
Stanislau-Snapok, line 185: In `defaultTestProps` we have only `initialValues`, but we redefine it below, so correct me if i wrong, but we don't need `defaultTestProps` here and in few places below.

PR #814, opened 2021-10-27
UICIRC-708: TLR setting cannot be disabled when there is an active title level request in the system

Stanislau-Snapok:
## Purpose
Title level request setting cannot be disabled when there is an active title level request in the system.

## Approach
Now we have TLR checkbox change handler. If we want disable TLR function, first of all we check if atleast one TLR in system, and if it is - we show modal with warning and didn't change the state of checkbox.

Also this PR fix unexpected problem with react-final-form when library replace `null` value. Now we handle this in `getInitialValues` and `normalizeData`  functions, and change `null` value on one of acceptable for react-final-form

## Refs
https://issues.folio.org/browse/UICIRC-708


Stanislau-Snapok:
@zburke, @mkuklis, could you please review?

translations/ui-circulation/en.json
```diff
@@ -304,6 +304,8 @@
   "settings.titleLevelRequests.expirationNotice": "Expiration notice",
   "settings.titleLevelRequests.desription": "Select templates for the following patron notices. If no template is selected, no notice will be sent.",
   "settings.titleLevelRequests.defaultTemplateValue": "Select notice",
+  "settings.titleLevelRequests.forbiddenDisableTlrModal.title": "Cannot change \"Allow title level requests\"",
+  "settings.titleLevelRequests.forbiddenDisableTlrModal.description": "\"Allow title requests\" cannot be changed because it is in use by one or more requests",
```
zburke, line 308: "Allow title level" => "Allow title-level"
"Allow title requests" => "Allow title-level requests"

zburke (approved):
Code-wise, everything looks great! This is a big feature and it's nicely implemented here. Nice job!

But the copy editor in me wants to tweak the language in the modal so its title is consistent with its text (title level requests vs title requests) and I think title-level should be hyphenated. Obviously, these are not deal-breakers. Since the wording here will be used as the basis for other translations, however, it's important that we get it correct, so if you could review it with the PO, I would appreciate it!

mkuklis (approved):
Nice work @Stanislau-Snapok!

translations/ui-circulation/en.json
Stanislau-Snapok: Thank you! Unfortunately my PO is on vacation for the rest of the week. But I promise to discuss it with her when she returns. I must resolve this conversation now, because other way this task will be a spillover for our team. Hope you have no any objections. Sorry for the rush and thank you again.

PR #815, opened 2021-10-29
UICIRC-709: Add possible for resizing panel in circulation rules editor

Dmitriy-Litvinenko:
## Purpose
Add possible for resizing panel in circulation rules editor

## Refs
https://issues.folio.org/browse/UICIRC-709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Locations display as `{code} {name}` and thus should be sorted by code, name. Refs UICIRC-703.
 * Add RTL/Jest testing for `PatronNoticeDetail` component in `settings/PatronNotices`. Refs UICIRC-643.
 * Title level request setting cannot be disabled when there is an active title level request in the system. Refs UICIRC-708.
+* Add possible for resizing panel in circulation rules editor. Refs UICIRC-709.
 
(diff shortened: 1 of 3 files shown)


zburke (commented):
I don't understand this code well enough (or why this fix works) to feel comfortable approving it. @JohnC-80 , can you take a look?

Dmitriy-Litvinenko:
Hello @zburke sorry i should to add this description from the beginning
I investigated this problem. Result of investigation below:
1. How i understood `<Paneset>` used for wrap several `<Pane>` (https://ux.folio.org/storybook/?path=/story/pane--basic-usage) Just for example:
```
<Paneset>
  <Pane>
    Pane Content 1
  </Pane>
  <Pane>
    Pane Content 2
  </Pane>
</Paneset>
```
2. I checked our layout and found an extra `<Paneset>` and i deleted it
image

zburke:
Hmmm; [nested `<Panesets>` are specifically described in the documentation though](https://github.com/folio-org/stripes-components/tree/17a429ebd8f5ff38e4a42d2a190863f37456faa7/lib/Paneset#usage). I mean, I'm glad that un-nesting them here fixes a layout glitch, but it feels like there's maybe a stripes-components bug that remains unsolved. Agreed, @JohnC-80?

zburke (approved):

PR #818, opened 2021-11-06
UICIRC-669: prevent deletion of policies if they're in use

Baroquem:
https://issues.folio.org/browse/UICIRC-669

This adds a HOC to provide the strings and functions necessary to check for policy use and display error messages, as that behavior is needed for loan policies, overdue fine policies, and lost-item fee policies. The actual mechanics of the check are handled by `<EntryManager>`, but it needs to be provided with the proper props.


zburke (approved):
This looks great @Baroquem! I wish there were a better way to pass the UUID of the policy into `withPreventDelete::updateSelectedPolicy()` than sniffing the path, but given the way I wrote the routing in `<EntryManager>` all those years ago, I'm not sure there is 😞 . That's my only reservation; the rest of this, especially the comments, is awesome!

src/settings/wrappers/withPreventDelete.js
```diff
@@ -0,0 +1,96 @@
...
+    }).isRequired,
+    resources: PropTypes.shape({
+      loans: PropTypes.object,
+    }).isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.updateSelectedPolicy = this.updateSelectedPolicy.bind(this);
```
mkuklis, line 52: @Baroquem I think we can avoid binding by changing the signature of `updateSelectedPolicy` (similar to what you've done with `isPolicyInUse`). `updateSelectedPolicy = () => {}`

mkuklis (approved):
@Baroquem great work! 💯

src/settings/wrappers/withPreventDelete.js
Baroquem: Good catch!

Baroquem:
> I wish there were a better way to pass the UUID of the policy into `withPreventDelete::updateSelectedPolicy()` than sniffing the path, but given the way I wrote the routing in `<EntryManager>` all those years ago, I'm not sure there is 😞 . 

Yes, that's the bit I wasn't too happy with. I did try for a cleaner solution, but I couldn't make it work.

PR #835, opened 2021-12-13
UICIRC-480: UI fee/fine date/time token previews are hard-coded

Stanislau-Snapok:
## Purpose
UI fee/fine date/time token previews are hard-coded.

## Approach
Now we set preview values depends on tenant locale. This locale we get from `intl.locale`.
Below I attached two screenshots for different locales. 

## Refs
https://issues.folio.org/browse/UICIRC-480

## Screenshots
For en-US locale:
image

For en-GB locale:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 * Add RTL/Jest testing for `ExceptionCard` component in `settings/LoanHistory/ExceptionCard`. Refs UICIRC- 604.
 * Add RTL/Jest testing for `FixedDueDateScheduleDetail` component in `settings\FixedDueDateSchedule`. Refs UICIRC-608.
 * Add RTL/Jest testing for `RequestPolicyDetail` component in `src\settings\RequestPolicy`. Refs UICIRC-646.
+* UI fee/fine date/time token previews are hard-coded. Refs UICIRC-480.
 
 ## [6.0.0] (https://github.com/folio-org/ui-circulation/tree/v6.0.0) (2021-09-30)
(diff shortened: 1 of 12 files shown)


Dmitriy-Litvinenko (approved):

mkuklis (approved):
Awesome work @Stanislau-Snapok!

Stanislau-Snapok:
@zburke Could you please review too?
Thanks!

zburke (approved):
It's great to see these values being localized in the templates; this will be a lot more clear for folks outside the en-us locale! 

I'm curious whether you investigated using `<FormattedDate>` and `<FormattedTime>` directly form `react-intl`, as [indicated in the i18n documentation](https://github.com/folio-org/stripes/blob/master/doc/i18n.md#dates-and-times), rather than pulling out the locale and feeding it the moment-based formatter. The formatter you implemented here works fine, but it is not how we typically handle date and time data.

Stanislau-Snapok:
> It's great to see these values being localized in the templates; this will be a lot more clear for folks outside the en-us locale!
> 
> I'm curious whether you investigated using `<FormattedDate>` and `<FormattedTime>` directly form `react-intl`, as [indicated in the i18n documentation](https://github.com/folio-org/stripes/blob/master/doc/i18n.md#dates-and-times), rather than pulling out the locale and feeding it the moment-based formatter. The formatter you implemented here works fine, but it is not how we typically handle date and time data.

I checked your suggestion and found that if we use `<FormattedDate>` the template values just disapear at all. I think it's connected with value types. Here values was hardcoded with `string` type, `moment` after formating returns `string` value too.
Thanks for review!

PR #890, opened 2022-03-18
UICIRC-768: Fix problem with module launch

Stanislau-Snapok:
## Purpose
We have a problem with module launch within development process.

## Approach
We had some cyclic dependencies  before. This PR resolves the problem. Also fixed dependent tests.

## Refs
https://issues.folio.org/browse/UICIRC-768


zburke (approved):
LGTM!

PR #908, opened 2022-04-29
UICIRC-784: fix the issue when some values in `overdue fine policy` could be less than 0

Stanislau-Snapok:
## Purpose
User must not be able to save form if "Overdue fine" or "Overdue recall fine" have values less than 0. `isFloatGreaterThanOrEqualToZero` validator must be added to these fields.

## Refs
https://issues.folio.org/browse/UICIRC-784

## Screenshots
before:
image

after:
image

### DIFF
--- a/src/settings/Models/FinePolicy/FinePolicy.js
+++ b/src/settings/Models/FinePolicy/FinePolicy.js
@@ -33,6 +33,11 @@ export default class FinePolicy {
 
   hasValue(pathToValue) {
     const value = get(this, pathToValue);
+    return value > 0;
+  }
+  
+  hasNonZeroValue(pathToValue) {
+    const value = get(this, pathToValue);
 
     return !!parseFloat(value, 10);
   }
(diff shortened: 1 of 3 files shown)


artem-blazhko (approved):

src/settings/Models/FinePolicy/FinePolicy.js
```diff
@@ -33,7 +33,8 @@ export default class FinePolicy {
 
   hasValue(pathToValue) {
     const value = get(this, pathToValue);
-    return value > 0;
+
+    return !!parseFloat(value, 10);
```
zburke, line 37: `parseValue("0.00")` will return false. Given it used to reject 0 or any negative value this may not be wrong necessarily, but it means the title remains misleading, i.e. previously it should have been `hasPositiveValue` and now it should be `hasNonZeroValue`. 

This change seems disconnected from the description in the PR. What is the goal with this change?

zburke (commented):
The `fines.js` changes look good but I don't understand the `FinePolicy.js` changes. Please either explain why that change is relevant. Or, if it's a separate issue, move it to a separate PR.

zburke (approved):
Thanks for the explanation. LGTM!

PR #915, opened 2022-05-23
UICIRC-782: No validation for spacebar when create Patron notice template

artem-blazhko:
## Purpose
Add additional validation for spacebar when create Patron notice template

## Refs
[UICIRC-782](https://issues.folio.org/browse/UICIRC-782)


Stanislau-Snapok (approved):

src/settings/Validation/engine/handlers.js
```diff
@@ -14,6 +14,10 @@ export const isNotEmpty = (value) => {
   return isNumber(value) || !isEmpty(value);
 };
 
+export const isNotEmptyWithoutSpace = (value) => {
+  return typeof value === 'string' && value.trim();
```
pavelspichonak, line 18: nit: => `&& !!value.trim()`

pavelspichonak (approved):

src/settings/Validation/engine/handlers.js
artem-blazhko: I am not sure that it is necessary because we have implicit type conversion with `&&` operator

artem-blazhko: fixed

mkuklis (approved):

PR #923, opened 2022-05-25
UICIRC-815: Update NodeJS to v16 in GitHub Actions

Dmitriy-Litvinenko:
## Purpose
Update NodeJS to v16 in GitHub Actions

## Refs
https://issues.folio.org/browse/UICIRC-815

#### TODOS and Open Questions
Task was created for investigate problem in test https://issues.folio.org/browse/UICIRC-817


zburke (approved):
LGTM! 

WRT that one failing test: we had similar problems in ui-users where it appears there were some lurking NPEs due to unsafe/uncareful parsing of responses, e.g. `foo.bar...` should really be `foo?.bar...` Here, it looks like an empty response from the query in `getScheduleByName` (i.e. one that doesn't contain `fixedDueDateSchedules`) would cause that. Maybe `fetch` isn't being mocked in the test? Good luck! And don't fret about one failing test; see folio-org/ui-users/pull/2144 for some real carnage 😱 !

PR #971, opened 2022-11-17
Fix problem with validate unique name

Dmitriy-Litvinenko:
## Purpose
Fix problem with validate unique name for:
- Patron notice templates
- Request policies
- Fixed due date schedules

## Approach
We should not render component until all necessary data will be loaded on layer edit.
Moved validate name by id to utilities and reuse it in all places.

## Refs
https://issues.folio.org/browse/UICIRC-870
https://issues.folio.org/browse/UICIRC-872
https://issues.folio.org/browse/UICIRC-873

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change history for ui-circulation
 
+## IN PROGRESS
+* Fix problem with saving Notice template. UICIRC-870.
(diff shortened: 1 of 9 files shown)


artem-blazhko (approved):

Dmitriy-Litvinenko:
Hello @zburke, @mkuklis Could you please review?

vashjs (approved):

PR #973, opened 2022-12-01
UICIRC-876: Permission "Can create, edit and remove staff slips" : Cannot edit Description

artem-blazhko:
## Purpose
When a user has "Can create, edit and remove staff slips" permission he/she can not edit a "Description" section of Staff slip.

## Approach
The issue happened because the "Description" field relies on "settings.organizations.enabled" permission. 
In this case, I created new permission "ui-circulation.settings.edit-staff-slips" that will be used instead of "settings.organizations.enabled" and also will be a sub-permission of "ui-circulation.settings.staff-slips" permission.

## Refs
[UICIRC-876](https://issues.folio.org/browse/UICIRC-876)


Dmitriy-Litvinenko (approved):

zburke (changes requested):
Leveraging `settings.organization.enabled` makes ui-circulation dependent on `ui-organizations` without explicitly stating this dependency. That is wrong. If some of the same permissions needed by ui-organizations are needed by ui-circulation, so be it; they should be included in ui-circulation. 

LMK if you need me to follow up with the PO on the ticket.

artem-blazhko:
Hello @zburke.
We have explicit dependency on `ui-organizations` for a long time. See [this piece of code](https://github.com/folio-org/ui-circulation/blob/5278fbdb0fecaeac754bbf29b33260ce2fbe0485/src/settings/StaffSlips/StaffSlipForm.js#L61).
And we do not know what was the reason to use `'settings.organizations.enabled'` permission. In this case we decided to extend `'ui-circulation.settings.staff-slips'` permission (it was discussed with Julie Bickle).
Could you help to find a solution if my approach is not correct?
Thank you in advance.

zburke:
@artem-blazhko, my point is, do you really want ui-circulation to be dependent on ui-organizations, or is it the case that some operation in ui-circulation needs the same permissions as some operation in ui-organizations? Thinking of this another way, say the requirements to get a drivers license in France are the same as they are in Span. Should France say, "The requirements for getting a French license are 'go get a Spanish license'"? Or should France say, "The requirements for getting a French license are the same as for getting a Spanish license?"

Maybe it is the former, e.g. maybe ui-inventory should not show a link to an item's open loan (a page which lives within ui-users) if the user doesn't have `ui-users.can-view-loan-details`. Or maybe it is the latter, e.g. maybe ui-inventory has its own loan-details page and so it should contain a permission set like `ui-inventory.can-view-loan-details` which matches `ui-users.can-view-loan-details`.

artem-blazhko:
@zburke how can we handle this situation?
Should I create a separate permission on `ui-circulation` side that will be similar to `ui-organizations` permission?
If yes, I think it should be discussed with PO because I do not have a clear understanding of why `settings.organizations.enabled` was used initially.

zburke:
@artem-blazhko, I think we should follow up with the PO for guidance. I'll add a comment on the ticket and tag Cheryl Malmborg and Dennis Bridges, the POs for ui-circulation and ui-organizations, respectively.

zburke (approved):
LGTM

PR #974, opened 2022-12-01
UICIRC-874: The "Delete" button looks inconsistent on the "Fixed due date schedules" edit page

artem-blazhko:
## Purpose
"Fixed due date schedules" section of settings had inconsistent view of Edit page as compared with other settings. It has a Delete button on the bottom of the page but it should be in the "Actions" menu.

## Approach
Removed unnecessary functionality, tests, translations. Instead of that use delete functionality provided by `EntryManager` component.

## Refs
[UICIRC-874](https://issues.folio.org/browse/UICIRC-874)

### Before
image

image

### After
image


zburke (approved):
LGTM

Dmitriy-Litvinenko (approved):

PR #984, opened 2023-01-17
UICIRC-591: Remove bigtest tests and associated dependencies

Dmitriy-Litvinenko:
## Purpose
Remove bigtest tests and associated dependencies

## Refs
https://issues.folio.org/browse/UICIRC-591

## TODOS and Open Questions
1. How we understand current changes do not affect build-npm.yml and build-npm-release.yml and changes for this files will be merged separately if needed. Could you please notify me if i am wrong?
2. Could you please notify me if i am miss something or remove that used not only for bigtest?
3. We continue increase code coverage for RTL/Jest.


Dmitriy-Litvinenko:
Hello @zburke @mkuklis @artem-blazhko Could you please review when you will have time?
## TODOS and Open Questions
1. How we understand current changes do not affect build-npm.yml and build-npm-release.yml and changes for this files will be merged separately if needed. Could you please notify me if i am wrong?
2. Could you please notify me if i am miss something or remove that used not only for bigtest?
3. We continue increase code coverage for RTL/Jest.

zburke (changes requested):
OMG, `test/ui-testing`! Those are NightmareJS tests, probably even cruftier than BigTest! It feels good to let go of this baggage, doesn't it? 

As to `.gitignore` changes: I think  `.stripesclirc` should remain in `.gitignore`; it is a local config file that should never be committed. Removing the `.stripesclirc.js` file is correct, and if we need to add it to `.gitignore` to keep it from being added accidentally that would be fine.

mkuklis (approved):

Dmitriy-Litvinenko:
> OMG, `test/ui-testing`! Those are NightmareJS tests, probably even cruftier than BigTest! It feels good to let go of this baggage, doesn't it?
> 
> As to `.gitignore` changes: I think `.stripesclirc` should remain in `.gitignore`; it is a local config file that should never be committed. Removing the `.stripesclirc.js` file is correct, and if we need to add it to `.gitignore` to keep it from being added accidentally that would be fine.

Thank you @zburke Restore .stripesclirc in .gitignore ( https://github.com/folio-org/ui-circulation/pull/984/commits/7ca333ee33826ebe091be576b7ea1693fc8cd986 )

artem-blazhko (approved):

zburke (approved):
LGTM!

PR #989, opened 2023-01-23
UICIRC-892 Drop react-redux

ryandberger:
- Refs https://issues.folio.org/browse/UICIRC-892.
- Dropping unneeded `react-redux` from this component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change history for ui-circulation
 
-## (8.0.0) IN PROGRESS
+## IN PROGRESS
 * Make consistent view of delete functionality of "Fixed due date schedules". Refs UICIRC-874.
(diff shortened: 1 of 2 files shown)


Dmitriy-Litvinenko (changes requested):
Could you please clarify should we update dependency ("react-redux") or this is obsolete dependency and can be removed?

ryandberger:
> Could you please clarify should we update dependency ("react-redux") or this is obsolete dependency and can be removed?

I will try running tests without `react-redux` to determine if it can be dropped.

ryandberger:
> > Could you please clarify should we update dependency ("react-redux") or this is obsolete dependency and can be removed?
> 
> I will try running tests without `react-redux` to determine if it can be dropped.

Confirmed it is not needed.

zburke (approved):

PR #1033, opened 2023-05-18
Mark legacy permission as legacy in translations

MikeTaylor:
Also, update use of legacy permission as a sub-permission.


translations/ui-circulation/en_US.json
```diff
@@ -401,7 +401,7 @@
     "permission.settings.loan-history": "Settings (Circ): Can view loan history",
     "permission.settings.loan-policies": "Settings (Circ): Can create, edit and remove loan policies",
     "permission.settings.notice-policies": "Settings (Circ): Can create, edit and remove notice policies",
-    "permission.settings.circulation-rules": "Settings (Circ): Can create, edit and remove circulation rules",
+    "permission.settings.circulation-rules": "Settings (Circ): Can create, edit and remove circulation rules [LEGACY]",
```
MikeTaylor, line 404: I copied this changed translation through to `en_US`, because otherwise the change in `en` is not visible. That's not necessary for the two translations that I added in `en`, as new translations shine through the `en_US` overlay.

mkuklis (approved):

PR #1036, opened 2023-05-26
UICIRC-931: Add "Reminder fees" sub-section to Overdue Fine Policy

mkuklis:
## Purpose

Add "Reminder fees" sub-section to Overdue Fine Policy

## Refs

https://issues.folio.org/browse/UICIRC-933

## Screenshots

reminder-fees-details-view

reminder-fees-edit


artem-blazhko (commented):
Do we really need to update `en_US.json` file directly?
Could you check if we can do something with code smells?

mkuklis:
@artem-blazhko sometimes the English translations are delayed. I want to make sure they are available right away. It doesn't cause any issues with the incoming translations in the future.

mkuklis:
@artem-blazhko the code smells reported by the sonar cloud are not very useful under this PR. I need to rely on these classes to follow the existing validation patterns established earlier.

There are also similar cases already in the source code for example:

https://github.com/folio-org/ui-circulation/blob/master/src/settings/Models/FixedDueDateSchedule/Schedule.js#L1

MikeTaylor:
See also: https://sauropods.win/@mike/110547157079846685

MikeTaylor:
I suspect you're misreading the translations issue, though, @mkuklis, unless there is something else going on that I don't know about.

In my experience, newly _added_ translations shine through from `en.json` to all the other locales as soon as they're included in a build. It's when you _change_ the value of an existing translation in `en.json` that a "more specific" translation in `en_US.json` — really just a blind copy in 99% of cases — wins out until the next translations update.

So I don't think you needed to include `en_US.json` in this PR after all; but including it is certainly harmless.

mkuklis:
@MikeTaylor I was running into an issue with the translations when running it via:

https://orchid-dev2.folio-dev.indexdata.com

and only after I added them to en_US they started showing up correctly.

Maybe something is off in that environment. I just removed them from en_US as you both suggested.

mkuklis:
@artem-blazhko I made couple more small changes and the code smell issue should be now gone.

artem-blazhko (approved):

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 * Leverage cookie-based authentication in all API requests. Refs UICIRC-881.
 * Also support `circulation` `14.0`. Refs UICIRC-935.
 * Add request.requestDate as staff slip token in Settings. Ref CIRC-1784
+* Add "Reminder fees" sub-section to Overdue Fine Policy. Refs UICIRC-931.
```
artem-blazhko, line 29: Could you move it to the end of the list?

PR #1053, opened 2023-07-11
UICIRC-955: Add additional fields to reminder fees

mkuklis:
https://issues.folio.org/browse/UICIRC-955

Add additional fields to the reminder fees section.

### DIFF
--- a/src/settings/FinePolicy/components/EditSections/ReminderFeesSection/ReminderFeesSection.js
+++ b/src/settings/FinePolicy/components/EditSections/ReminderFeesSection/ReminderFeesSection.js
@@ -55,6 +55,28 @@ const ReminderFeesSection = ({ sectionOpen, noticeTemplates, blockTemplates }) =
         <Row>
           <Col xs={3}>
             <Field
+              data-testid="ignoreGracePeriodHoldsTestId"
+              label={<FormattedMessage id="ui-circulation.settings.finePolicy.reminderFees.ignoreGracePeriodHolds" />}
+              name="reminderFeesPolicy.ignoreGracePeriodHolds"
+              dataOptions={dataOptions}
+              component={Select}
+            />
+          </Col>
+        </Row>
+        <Row>
+          <Col xs={3}>
+            <Field
+              data-testid="allowRenewalOfItemsWithReminderFeesTestId"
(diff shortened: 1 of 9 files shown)


src/settings/FinePolicy/components/EditSections/ReminderFeesSection/ReminderFeesSection.js
```diff
@@ -52,6 +52,28 @@ const ReminderFeesSection = ({ sectionOpen, noticeTemplates, blockTemplates }) =
             />
           </Col>
         </Row>
+        <Row>
+          <Col xs={3}>
+            <Field
+              data-testid="ignoreGracePeriodHoldsTestId"
+              label={<FormattedMessage id="ui-circulation.settings.finePolicy.reminderFees.ignoreGracePeriodHolds" />}
+              name="reminderFeesPolicy.ignoreGracePeriodHolds"
+              dataOptions={dataOptions}
```
MikeTaylor, line 61: As a tiny tiny matter of style, I prefer to say `component={Select}` _before_ `dataOptions={dataOptions}`, as the data-options are a refinement of how the selector works. Feel free to ignore this if you don't agree, or if consistent style in this code is to use the order you went with here.

src/settings/Models/ReminderFeesPolicy/ReminderFeesPolicy.test.js
```diff
@@ -23,12 +23,26 @@ describe('ReminderFeesPolicy', () => {
 
   it('"defaultReminderFeesPolicy" should return correct data', () => {
     const expectedPolicy = {
-      countClosed: true,
-      ignoreGracePeriodRecall: true,
-      clearPatronBlockWhenPaid: true,
+      countClosed: false,
```
MikeTaylor, line 26: Is it deliberate that the values provided for some of the existing field have changed?

mkuklis: Yes @MikeTaylor, the previous implementation was buggy. What is coming by default from the server is `false` so we just need to make sure we align with that here.

src/settings/FinePolicy/components/EditSections/ReminderFeesSection/ReminderFeesSection.js
mkuklis: good idea @MikeTaylor!

PR #1054, opened 2023-07-14
UICIRC-938: Adjust address token names

mkuklis:
https://issues.folio.org/browse/UICIRC-938

The work for this was done in https://github.com/folio-org/ui-circulation/pull/1040
This PR just adjusts the token names based on the feedback from PO.


MikeTaylor (approved):
Much better :-)

PR #1080, opened 2023-10-11
UICIRC-992: Also support feesfines interface version 19.0

Dmitriy-Litvinenko:
## Purpose
Also support `feesfines` interface version `19.0`

## Approach
Changes was in endpoint `/accounts` that do not used in current module
This is part of `Quesnelia (R1 2024)` release
Should be merged after merging https://issues.folio.org/browse/MODFEE-337

## Refs
https://issues.folio.org/browse/UICIRC-992

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change history for ui-circulation
 
-## 10.0.0 IN PROGRESS
+## 9.1.0 IN PROGRESS
 
 * Also support `feesfines` interface version `19.0`. Refs UICIRC-992.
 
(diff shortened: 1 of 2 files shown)


zburke (changes requested):
The new interface support looks fine, but assuming this is supposed to land if v9, you need to fix the CHANGELOG entry. Or if this _is_ intended for v10 then you need to bump the version in `package.json` too.

Dmitriy-Litvinenko:
@zburke Could you please review (package.json updated)? Current pull request will be merged after release previous release.

zburke (approved):
👌 Adding a new version of an existing interface is fine in a minor version. The thing that's tricky is removing supported versions, e.g. changing from 
```
"feesfines": "16.2 17.0 18.0"
```
to
```
"feesfines": "16.3 17.0 18.0"
```
is a breaking change because you're removing support for something you used to handle.

artem-blazhko (approved):

PR #1122, opened 2024-02-14
UICIRC-1047: RulesEditor should use stripes.hasPermission

Dmitriy-Litvinenko:


zburke (approved):
❤️ that was fast; thank you!

PR #1130, opened 2024-03-12
UICIRC-1064: Fix that Save & close button is missing in the circulation forms

Dmitriy-Litvinenko:
## Purpose
Fix that Save & close button is missing in the circulation forms

# Refs
https://issues.folio.org/browse/UICIRC-1064

### DIFF
--- a/src/settings/FinePolicy/FinePolicyForm.js
+++ b/src/settings/FinePolicy/FinePolicyForm.js
@@ -80,8 +80,7 @@ class FinePolicyForm extends React.Component {
       : formatMessage({ id: 'ui-circulation.settings.finePolicy.createEntryLabel' });
 
     const footerPaneProps = {
-      pristine,
-      submitting,
+      isSaveButtonDisabled: pristine || submitting,
       onCancel,
     };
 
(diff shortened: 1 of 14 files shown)


Terala-Priyanka (approved):

src/settings/FinePolicy/FinePolicyForm.test.js
```diff
@@ -68,7 +68,7 @@ describe('FinePolicyForm', () => {
   const policyForTest = new FinePolicy(mockedInitialValues);
   const defaultProps = {
     stripes: mockedStripes,
-    pristine: true,
+    pristine: false,
```
Dmytro-Melnyshyn, line 71: The default value of pristine in a `Form` is `true`, which means the form values are not changed. It is not clear why it becomes false.

Dmytro-Melnyshyn (changes requested):

src/settings/FinePolicy/FinePolicyForm.js
```diff
@@ -80,8 +80,7 @@ class FinePolicyForm extends React.Component {
       : formatMessage({ id: 'ui-circulation.settings.finePolicy.createEntryLabel' });
 
     const footerPaneProps = {
-      pristine,
-      submitting,
+      isSaveButtonDisabled: pristine || submitting,
```
Dmytro-Melnyshyn, line 83: The synonym of `pristine` is untouched it is strange that the button is disabled when the state of a `Form` is untouched.

src/settings/FinePolicy/FinePolicyForm.test.js
Dmitriy-Litvinenko: Fixed

Dmytro-Melnyshyn (approved):

PR #1174, opened 2024-11-12
UICIRC-1122: Fix DOMPurify import

Dmitriy-Litvinenko:
## Purpose
Fix DOMPurify import

## Refs
https://issues.folio.org/browse/UICHKIN-435

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/circulation",
-  "version": "9.2.0",
+  "version": "10.0.0",
   "description": "Circulation manager",
   "repository": "folio-org/ui-circulation",
   "publishConfig": {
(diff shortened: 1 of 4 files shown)


artem-blazhko (approved):

artem-blazhko (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@folio/circulation",
-  "version": "9.2.0",
+  "version": "10.0.0",
```
zburke, line 3: This does not represent a breaking change. It is customary to label breaking changes in the CHANGELOG so others can see what changes caused the version to bump. You don't _have_ to change this here if you know the next release will require a new major-version, but it is a little odd that this isn't a breaking change, and there isn't an explanatory comment.

zburke (approved):

PR #1183, opened 2024-12-03
UICIRC-1128: Added missing translations

Dmitriy-Litvinenko:
## Purpose
Added missing translations

## Refs
https://folio-org.atlassian.net/browse/UICIRC-1128


zburke (commented):
Use the PR description to change what is changing, i.e. what the state of the code will be after this change is complete. If you're fixing a bug, don't describe the bug. After the PR merges, the bug will be gone, so a description of a problem that used to be there isn't important. 

This PR adds several new translation keys but does not update any code to point to them. After this PR merges, nothing will change. Am I missing something? Were these keys missing? If so, state that in the description, e.g. instead of "incorrect name for TLR" say "Provide missing translations".

Dmitriy-Litvinenko:
> Use the PR description to change what is changing, i.e. what the state of the code will be after this change is complete. If you're fixing a bug, don't describe the bug. After the PR merges, the bug will be gone, so a description of a problem that used to be there isn't important.
> 
> This PR adds several new translation keys but does not update any code to point to them. After this PR merges, nothing will change. Am I missing something? Were these keys missing? If so, state that in the description, e.g. instead of "incorrect name for TLR" say "Provide missing translations".

Thank you rename to "Added missing translations"

artem-blazhko (approved):

zburke (approved):

PR #1216, opened 2025-04-17
UICIRC-1194: Replace moment with day.js

artem-blazhko:
## Purpose
 Replace moment with day.js

## Refs
[UICIRC-1194](https://folio-org.atlassian.net/browse/UICIRC-1194)


zburke (approved):
Changes LGTM. Overall, I question the need for much of the date-parsing that happens in this module, but without any documentation or examples, it is difficult to assess whether we truly need dayjs/moment or if in fact react-intl provides everything we need.

Dmitriy-Litvinenko (approved):

